PR #2, opened 2015-10-26
Add support for embedded JSON-LD

redapple:


redapple:
@eliasdorneles ,
what do you think?

eliasdorneles:
Hey @redapple, nice feature, looks good!

One thing I noticed is that `extract_items` here returns a list of items (which seems fine) but the microdata extractor returns a dict with an `items` key, which seems a bit inconsistent.

Is this the expected behavior?
Wouldn't it make sense to change the microdata extractor to return a list too?

Keeping the same interface across extractors would probably make it easier for the service (which could be updated to use this one too, maybe in a separated PR).

redapple:
Indeed,
`w3cmicrodata` follow http://www.w3.org/TR/microdata/#json, which defines the results as an array containing an `items` field.
For JSON-LD, I haven't found the equivalent guidelines for the result format.
But it does make sense to return lists in all cases.

redapple:
I used `{"items": ...}` as well, to leave room for additional metadata

eliasdorneles:
LGTM! :+1:

PR #11, opened 2016-04-07
Change service response content_type to JSON

stummjr:
The service is currently returning JSON content as `text/html` and I changed it to `application/json`.

I use a browser extension that highlights syntax when viewing source code files and it wasn't rendering JSON properly because of the response content type. :-)


redapple:
Thanks @stummjr !

PR #26, opened 2016-09-29
RDFa extractor based on rdflib's pyRdfa and backed by lxml

redapple:
This RDFa parser uses lxml instead of html5lib.
It uses custom element classes for lxml so that the custom lxml parser generates nodes that are compatible with `xml.dom`, that the original pyRdfa expects.

Warning: there are no real tests yet, and performance has not be tested.


legel:
Hello, this looks like a very exciting addition, although it's unclear why the tests aren't passing for the new module? In any case, thanks for the contributions...

legel:
Thanks @redapple, looking forward. I'm working on a large-scale crawling project and have already begun to make a few changes that introduce a little more stability and parsing flexibility.  One thing I'm seeing in the big picture is that it would be very nice to have a JSON parser that extracts whole ontological objects regardless of the JSON structure, since so many sites ultimately don't follow specs and yet contain sufficient structure to dynamically parse into a uniform data structure.  I'll keep you posted.  I'm going to continue iterating and when I feel things are definitely okay all around will look to submit a pull request.

redapple:
@legel , looking forward to your contributions.
what do you mean by "JSON parser" here?

legel:
Thanks @redapple, looking very nice; also excited to see further development around the open graph tags that so many sites are using.  If I put something together in the extruct framework I'll let you know.

Meanwhile I probably spoke too loosely in the term "JSON parser" because all that I really meant (and am finding useful on my side) is using classes that correspond to the Schema classes, e.g. a class for Event (http://schema.org/Event) that has as member variables the values that are extracted from the raw HTML. The age-old challenge as you can imagine with scraping many different semantic sites is that so many of them use slightly different (incorrect) syntaxes and nesting of data, and so no matter there is a lot of parsing required to organize the useful information for a particular type of Schema in the face of unknown web page structures. This is just my experience but possibly there is something I'm doing here that could be of general use for all classes... i.e. cut through the most common types of mistakes people make in encoding schemas by automatically parsing them into standard object types.

redapple:
I'm merging this to get an alpha release out.
No guarantees whatsoever.

PR #33, opened 2017-01-09
Add PyPy3 build on Travis CI

redapple:


redapple:
Looks like it's not easy to fix: https://github.com/yyuu/pyenv-virtualenv/issues/188

PR #38, opened 2017-03-26
Add cloneNode and toxml as some pages fail when extracting RDFa

andrix:
Trying to extract some RDFa from a site, I got AttributeError exception because of the lack of implementation of cloneNode and toxml. 

I've added both methods and now it's working.


redapple:
Good catch @andrix !
I can reproduce the exceptions with this page for example: http://nielslubberman.nl/drupal/
It's related to `'http://purl.org/rss/1.0/modules/content/encoded'` property (with RDFa 1.0 at least)

PR #47, opened 2017-06-12
Support leading HTML or JavaScript comment in JSON-LD payload

redapple:


redapple:
@giordand  , would this fix work for you?
It only handles
```
<script type="application/ld+json">
  // John allocated 5 hours to exercise.
{
...
```
and
```
<script type="application/ld+json">
<!-- John joined the festival. -->
{
...
```

PR #50, opened 2017-06-19
Silence rdflib logger to avoid message when importing

stummjr:
`rdflib` displays this annoying log message when we import it. Thus, importing `extruct` gives us this:

    >>> import extruct
    INFO:rdflib:RDFLib Version: 4.2.2

This message is also displayed when we run `python -m extruct` and it's irrelevant to extruct, so this PR silences the `rdflib` logger before importing that module.


redapple:
Thank you @stummjr

PR #51, opened 2017-06-19
Add support for 'python -m extruct' command

### LINKED ISSUE #46: Allow calling extruct cmd line tool using "python -m extruct"
It would be nice to be able to call the extruct command line tool using `python -m extruct`.

With this, people will be able to use a specific Python version and it may also help people with issues on their system's PATH.

stummjr:
Fixes #46.


redapple:
Nice @stummjr !

PR #52, opened 2017-06-20
Add release notes for v0.4.0

stummjr:


HISTORY.rst
```diff
@@ -2,6 +2,12 @@
 History
 =======
 
+v0.4.0 (2017-06-20)
+-------------------
+
+* New ``extruct`` command line tool to fetch a page and extract its metadata.
+* Add support for leading HTML comment in JSON-LD payload.
```
redapple: Maybe something like "Accept leading HTML comment in JSON-LD payload"

stummjr: Just updated it.

PR #56, opened 2017-11-10
Account for if jsonld is null

mikhuang:
I recently found a URL for which ld+json was set to null. Wasn't sure where else this should have been caught so changed at the point where failure occurs


extruct/jsonld.py
```diff
@@ -21,10 +22,12 @@ def extract(self, htmlstring, url='http://www.example.com/', encoding="UTF-8"):
         return self.extract_items(lxmldoc)
 
     def extract_items(self, document, *args, **kwargs):
-        return [item for items in map(self._extract_items,
-                                      self._xp_jsonld(document))
-                     for item in items
-                         if item]
+        try:
+            return [item for items in map(self._extract_items,
+                                          self._xp_jsonld(document))
```
redapple: There could be only one JSONLD that is empty in the page, and this would swallow the potentially succesful extruction.
`_extract_items` is probably a better place for this catch.

mikhuang:
Thanks for the feedback, it was very helpful! Not sure if additional commits trigger anything, submitted some tests and hopefully fixes a little bit ago.

extruct/jsonld.py
```diff
@@ -33,6 +33,11 @@ def _extract_items(self, node):
         except ValueError:
             # sometimes JSON-decoding errors are due to leading HTML or JavaScript comments
             data = json.loads(HTML_OR_JS_COMMENTLINE.sub('', script))
+
```
redapple: I'm tempted to suggest using generator like this:
```
    def _extract_items(self, node):
        script = node.xpath('string()')
        try:
            data = json.loads(script)
        except ValueError:
            # sometimes JSON-decoding errors are due to leading HTML or JavaScript comments
            data = json.loads(HTML_OR_JS_COMMENTLINE.sub('', script))

        if isinstance(data, list):
            for d in data:
                yield d
        elif isinstance(data, dict):
            yield data
```

wRAR (approved):

Issue #59, opened 2018-02-22
Memory leak issue


mahehss:
Steps to reproduce:
Login to Ubuntu box:
Run extruct as a service
nohup python -m extruct.service &

hit the http://localhost:10005/extruct/<url> in loop for 20k urls. Memory consumption increases with time and never comes down

Kdoggett887:
@mahehss I had the same issue when I was running extruct on multiple websites, so I forked the repository and was able to fix the issue. 

In the `DomHtmlElementClassLookup` class in the `xmldom.py` file, there is a class variable called `_lookups` which is a dict that stores information for parsing. `_lookups` stores keys/values as tuples/xmldom classes. An example entry looks like `(u'element', <lxml.etree._Document object at 0x103176ba8>, None, 'abbr'): <class 'extruct.xmldom.DomHtmlElement'>`. Since each document is part of the key, I clear `_lookups` before the metadata is returned from the `extract` method. In my case I am not processing documents more than once, so this solution shouldn't degrade performance.

This fixed the memory leak during a test. After processing about 240MB of files without the fix, usage was at about 1.4GB and with the fix usage was at about 130MB.

If this behavior is practical for all cases, I could do a pull request.

lopuhin:
@Kdoggett887 I think solution you described makes sense, please send a PR! Especially given that memory leak happens not in the server but in the extractors themselves.

redapple:
@Kdoggett887 , it's great that you could track this down. I was suspecting something around the parser that was introduced for RDFa, but hadn't had time to look into it. Awesome!

PR #60, opened 2018-03-13
Fix memory leak, issue-59

Kdoggett887:
Resets a class variable after each extraction to prevent memory leak. More details in #59 .

### DIFF
--- a/tests/test_extruct.py
+++ b/tests/test_extruct.py
@@ -3,7 +3,6 @@ import json
 import unittest
 
 import extruct
-from extruct.xmldom import DomHtmlElementClassLookup
 from tests import get_testdata
 
 
--- a/tests/test_extruct.py
+++ b/tests/test_extruct.py
@@ -17,4 +17,3 @@ class TestGeneric(unittest.TestCase):
 
         data = extruct.extract(body, 'http://www.songkick.com/artists/236156-elysian-fields')
         self.assertEqual(data, expected)
-        self.assertEqual(DomHtmlElementClassLookup._lookups, {})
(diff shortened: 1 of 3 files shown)


lopuhin:
@Kdoggett887 thanks for PR! Sorry if the answer to my question is obvious, as I'm new to extruct, but I wonder why is ``_lookups`` a class attribute instead of an instance attribute? As far as I understand, if it would be an instance attribute, then it would be cleaned up automatically since we create a new parser in ``extruct.extract`` function?

Kdoggett887:
@lopuhin Thanks for the suggestion! I was looking into doing this, but was unsuccessful at first and gave up on it fairly quickly since it caused some upstream issues in the lxml package which I'm unfamiliar with.

After your suggestion, I decided to look into it more and found a solution that puts the `_lookups` variable into the constructor. When I had tried it before, I wasn't instantiating the `HtmlElementClassLookup`, so the constructor just instantiates the super class and assigns `self._lookups = {}`. I'm testing the memory usage on a large sample of files now, but if it works this should be a better solution.

tests/test_extruct.py
```diff
@@ -3,6 +3,7 @@
 import unittest
 
 import extruct
+from extruct.xmldom import DomHtmlElementClassLookup
```
kmike: why is this import needed?

Kdoggett887:
@lopuhin thanks for the feedback! I fixed up the spacing and removed the unused import. I ran a test checking the speed and the updated version took ~85% of the time master did, but more testing would be needed to confirm the results.

kmike:
Awesome, thanks @Kdoggett887!

PR #61, opened 2018-03-14
Add og and microformat extraction

Kebniss:
Progresses so far:
- Added opengraph and microformat extraction 
- Unified parsing for rdfa, microdata, opengraph and jsonld. Microformat relies on an external lib that uses bs4 to parse so it's not as easy to integrate it as well


extruct/__init__.py
```diff
@@ -1,16 +1,107 @@
...
+from extruct.opengraph import OpenGraphExtractor
+from extruct.microformat import MicroformatExtractor
 from extruct.xmldom import XmlDomHTMLParser
 
 
-def extract(htmlstring, url='http://www.example.com/', encoding="UTF-8"):
+logger = logging.getLogger(__name__)
+
+
+def _microdata_transform(obj, schema_context, add_context=True):
```
kmike: I think it is better to keep top-level `__init__.py` small, it should contain imports and the highest-level extract function. I'd even move `extract` function elsewhere as well.

extruct/pathfinder.py
```diff
@@ -0,0 +1,62 @@
+import itertools
```
kmike: Is `pathfinder.py` module really needed? find function is imported in opengraph.py, but it is unused. I'd prefer to remove it.

extruct/opengraph.py
```diff
@@ -0,0 +1,100 @@
...
+from .pathfinder import find
+from .transform import clean_missing
+from .schemaorg import (
+    schemaorg,
+    schemaorg_nested,
+    schemaorg_instock,
+)
+
+
+def opengraph_availability(avail):
```
kmike: While transformation to schema.org syntax is useful, I'm not sure extruct is a right place for it. I think it makes sense to keep only extraction in `extruct`, and have a separate package for transforming everything to the same data format.

kmike:
👍 That's great to have more semantic markup types in extruct!

Could you please remove everything related to transformation of output though, and make it return raw data instead? It should make PR much smaller.

Before merging it needs a few other changes:

* README should mention new syntaxes;
* there should be tests;
* extruct.tool should be updated;
* extruct.service should be updated.

extruct/opengraph.py
```diff
@@ -0,0 +1,77 @@
...
+
+import functools
+from .transform import clean_missing
+from .schemaorg import (
+    schemaorg,
+    schemaorg_nested,
+    schemaorg_instock,
+)
+
+def opengraph_transform(opengraph_entries):
```
kmike: let's remove transformations from this PR

extruct/schemaorg.py
```diff
@@ -0,0 +1,31 @@
+"""Schema.org function helpers."""
```
kmike: It looks like this module, as well as  extruct.transform are not needed if we remove transformations from the PR.

extruct/__init__.py
```diff
@@ -1,16 +1,39 @@
...
+        processors.append(('microformat', MicroformatExtractor().extract_items))
+    if 'rdfa' in syntaxes:
+        processors.append(('rdfa', RDFaExtractor().extract_items))           
+
+    output = {}
+    for label, extract in processors:
+        try:
+            output[label] = [obj for obj in extract(document=tree, url=url, html=html)]
+        except Exception:
+            logger.exception("Failed to parse %s", url)
```
kmike: why are exceptions silenced?

extruct/__init__.py
```diff
@@ -1,16 +1,39 @@
+import logging
+import functools
```
kmike: this import is unused

tests/samples/songkick/elysianfields.json
```diff
@@ -1,115 +1 @@
...
..."Buy tickets for an upcoming Elysian Fields concert near you. List of all Elysian Fields tickets and tour dates for 2017."], ["og:url", "http://www.songkick.com/artists/236156-elysian-fields"], ["og:image", "http://images.sk-static.com/images/media/img/col4/20100330-103600-169450.jpg"]]}], "microformat": [], "rdfa": [{"@id": "http://www.songkick.com/artists/236156-elysian-fields", "al:ios:app_name": [{"@value": "Songkick Concerts"}], "al:ios:app_store_id": [{"@value": "438690886"}], "al:ios:url": [{"@value": "songkick://artists/236156-elysian-fields"}], "http://ogp.me/ns#description": [{"@value": "Buy tickets for an upcoming Elysian Fields concert near you. List of all Elysian Fields tickets and tour dates for 2017."}], "http://ogp.me/ns#image": [{"@value": "http://images.sk-static.com/images/media/img/col4/20100330-103600-169450.jpg"}], "http://ogp.me/ns#site_name": [{"@value": "Songkick"}], "http://ogp.me/ns#title": [{"@value": "Elysian Fields"}], "http://ogp.me/ns#type": [{"@value": "songkick-concerts:artist"}], "http://ogp.me/ns#url": [{"@value": "http://www.songkick.com/artists/236156-elysian-fields"}], "http://www.facebook.com/2008/fbmlapp_id": [{"@value": "308540029359"}]}]}
```
kmike: Why is indentation removed? It is hard to check a diff without indentation, and it is also hard to read a file.

README.rst
```diff
@@ -54,197 +59,107 @@ All-in-one extraction
...
-               'http://ogp.me/ns#title': [{'@value': 'Elysian Fields'}],
-               'http://ogp.me/ns#type': [{'@value': 'songkick-concerts:artist'}],
-               'http://ogp.me/ns#url': [{'@value': 'http://www.songkick.com/artists/236156-elysian-fields'}],
-               'http://www.facebook.com/2008/fbmlapp_id': [{'@value': '308540029359'}]}]}
-
-
-You can also use each extractor individually. See below.
-
-Microdata extraction
-++++++++++++++++++++
```
kmike: Why is this section removed?

Kebniss: I am still modifying the readme, I committed it by mistake

tests/samples/songkick/elysianfields.json
Kebniss: I had to modify these files to include new syntaxes and did not think to keep indentation

README.rst
Kebniss: I had to remove because by adding microformat and opengraph the output was different and I found it easier to copy paste the new version instead of modifying the single lines

extruct/tool.py
```diff
@@ -56,10 +67,23 @@ def main():
...
+        action='store_true',
+        default=False,
+        help='Extract opengraph metadata from the page.',
+    )
     args = parser.parse_args()
 
-    if any((args.microdata, args.jsonld, args.rdfa)):
-        metadata = metadata_from_url(args.url, args.microdata, args.jsonld, args.rdfa)
+    if any((args.microdata, args.jsonld, args.rdfa, args.microformat, args.opengraph)):
+        metadata = metadata_from_url(args.url, args.microdata, args.jsonld,
```
kmike: Could you please use keyword arguments here? It is very easy to make a mistake otherwise.

tests/test_rdfa.py
```diff
@@ -81,26 +81,3 @@ def test_w3c_rdfaprimer(self):
...
-    def test_wikipedia_xhtml_rdfa(self):
-        body = get_testdata('misc', 'Portfolio_Niels_Lubberman.html')
-        expected = json.loads(
-                get_testdata('misc', 'Portfolio_Niels_Lubberman.json'
-            ).decode('UTF-8'))
-
-        rdfae = RDFaExtractor()
-        data = rdfae.extract(body, url='http://nielslubberman.nl/drupal/')
-
-        self.assertJsonLDEqual(data, expected)
```
kmike: why are these tests removed?

extruct/tool.py
```diff
@@ -32,6 +35,14 @@ def metadata_from_url(url, microdata=True, jsonld=True, rdfa=True):
         rdfae = RDFaExtractor()
         result['rdfa'] = rdfae.extract_items(tree, resp.url)
 
+    if opengraph:
+        oge = OpenGraphExtractor()
+        result['opengraph'] = [obj for obj in oge.extract_items(tree, resp.url)]
+
+    if microformat:
+        mfmate = MicroformatExtractor()
+        result['microformat'] = [obj for obj in mfmate.extract_items(html=resp.content, url=resp.url)]
```
kmike: Could this code be simplified now, as there is a top-level extraction function which allows to choose which formats to use?

extruct/__init__.py
Kebniss: I missed this comment before. I copied this bit from dse-microdata and I thought it was useful to keep it because in production if there is an error with parsing one syntax for a specific domain you don't necessarily want extraction to crash on all pages. As Konstantin mentioned very often webpages are annotated with multiple syntaxes and information is duplicated so the fact that, say rdfa, crashes on a page does not mean we need to stop on it

extruct/tool.py
```diff
@@ -6,10 +6,13 @@
...
 from extruct.w3cmicrodata import MicrodataExtractor
+from extruct.opengraph import OpenGraphExtractor
+from extruct.microformat import MicroformatExtractor
 from extruct.xmldom import XmlDomHTMLParser
 
 
-def metadata_from_url(url, microdata=True, jsonld=True, rdfa=True):
+def metadata_from_url(url, microdata=True, jsonld=True, rdfa=True,
+                      microformat=True, opengraph=True):
     resp = requests.get(url, timeout=30)
```
Kebniss: @kmike, @cathalgarvey  should we add headers here? Iirc some domains ban requests that do not have one

extruct/__init__.py
kmike: Yeah, I agree it makes sense to provide a way to ignore errors, but I think it'd be good to have an option to give users an access to the outcome (probably enabled by default). A function argument like `errors="log" / "ignore", "strict"`, with "strict" being a default value, can work; this would be similar to Python's "encode" function. 

Another option is to set `output[label]` to the exception object in case of error, but I like it less, as it looks like a trick.

Kebniss:
I don't get what test I should add to make codecov happy :/

lopuhin:
> I don't get what test I should add to make codecov happy :/

It seems that it's not happy with tool.py coverage, because there was a change there, but it was not covered by tests, it's visible on the diff view: https://codecov.io/gh/scrapinghub/extruct/compare/c465e629c9e35cff08a703f6d2912c1c71c642ff...b153c0b1718f4541e693d846d37bdfbe929d775b/diff

In some cases it's quite hard to cover all the diff, and PRs can be merged without passing this check. Here maybe https://pypi.python.org/pypi/requests-mock can help to test the main function to make it covered.

Kebniss:
I am already testing using request_mock as well I added microformat and opengraph tests

extruct/microformat.py
```diff
@@ -0,0 +1,14 @@
...
+
+class MicroformatExtractor(object):
+
+    def extract(self, htmlstring, url='http://www.example.com/', encoding='UTF-8'):
+        return list(self.extract_items(htmlstring, url=url))
+
+    def extract_items(self, html, url, document=None):
+        for obj in mf2py.parse(html, html_parser="lxml", url=url)['items']:
+            if obj['type'] == 'h-entry':
+                for o in obj['children']:
```
lopuhin: As far as I understand, this will turn a nested structure into a flat one - is this something that we want to do? Any examples where this happens in practice?

extruct/tool.py
lopuhin: I think it's fine without headers, I'd prefer keeping tool.py as simple as possible.

extruct/service.py
```diff
@@ -20,9 +20,10 @@ def _decorated(*args, **kwargs):
     return _decorated
 
 
-def async_extruct(url, microdata=True, jsonld=True, rdfa=True):
+def async_extruct(url, microdata=True, jsonld=True, rdfa=True,
+                  opengraph=True, microformat=True):
     response.content_type = 'application/json'
-    result = metadata_from_url(url, microdata, jsonld, rdfa)
+    result = metadata_from_url(url, microdata, jsonld, rdfa, opengraph, microformat)
```
lopuhin: Would be good to use keyword arguments here as well, as it's easy to confuse the order.

extruct/tool.py
```diff
@@ -1,65 +1,27 @@
...
-    if any((args.microdata, args.jsonld, args.rdfa)):
-        metadata = metadata_from_url(args.url, args.microdata, args.jsonld, args.rdfa)
-    else:
-        metadata = metadata_from_url(args.url)
+    arg = parser.add_argument
+    arg('url', help='The target URL')
+    arg('--syntaxes',
+        default='all',
+        help='Either list of microdata syntaxes to use or "all" (syntaxes\
+             available [microdata, microformat, rdfa, opengraph, jsonld])')
```
lopuhin: As far as I understand, here ``args.syntaxes`` would always be a string, while other functions look like they expect either an ``"all"`` string or a list of syntax names. Right now it will probably still work with any reasonable string as well, because ``'microdata' in syntaxes`` works correctly for a string, but I think it's better to convert syntaxes to a list (it it's not default "all" value), and provide an example string in valid format in `--syntaxes`` help.

tests/samples/songkick/tovestyrke.json
```diff
@@ -1,137 +1,192 @@
...
-        {
-          "@value": "308540029359"
-        }
-      ]
-    }
-  ],
-  "status": "200 OK",
-  "url": "https://www.songkick.com/concerts/30166884-tove-styrke-at-hoxton-square-bar-and-kitchen"
-}
+    "url": "https://www.songkick.com/concerts/30166884-tove-styrke-at-hoxton-square-bar-and-kitchen",
```
lopuhin: Is it the same with only indentation changed, or are there any other changes?

lopuhin: Never mind, I see that new fields are added...

tests/test_rdfa.py
```diff
@@ -83,24 +83,24 @@ def test_w3c_rdfaprimer(self):
             self.assertJsonLDEqual(data, expected)
 
     def test_wikipedia_xhtml_rdfa(self):
-        fileprefix = 'xhtml+rdfa'
```
lopuhin: Why is indentation changed here and below?

tests/test_tool.py
```diff
@@ -76,8 +79,40 @@ def test_metadata_from_url_rdfa_only(self, mock_get):
         )
         mock_get.return_value = mock_response
 
-        data = metadata_from_url(self.url, microdata=False, jsonld=False, rdfa=True)
-        self.assertEqual(data, expected)
+        data = metadata_from_url(self.url, syntaxes='rdfa')
```
lopuhin: I'm not sure, but it seems to me that it would be better to use and support only one way of passing syntaxes (except for default "all"), as a list of valid syntax names, not as a string that contains them.

extruct/__init__.py
```diff
@@ -1,16 +1,38 @@
...
+logger = logging.getLogger(__name__)
+
+
+def extract(htmlstring, url='http://www.example.com/', encoding="UTF-8",
+            syntaxes="all", schema_context='http://schema.org'):
     domparser = XmlDomHTMLParser(encoding=encoding)
     tree = fromstring(htmlstring, parser=domparser)
-    return {name: extractor.extract_items(tree, url=url)
-            for name, extractor in (
-                ('json-ld', JsonLdExtractor()),
```
lopuhin: This breaks backwards compatibility in the public API - it was "json-ld" before and now is "jsonld", I think it's better to keep it backwards compatible unless there is a strong reason to change it.

lopuhin:
> I am already testing using request_mock as well I added microformat and opengraph tests

Great! I think it's complaining because main function is not covered - could you please add test for it as well, especially since it contains non-trivial changes?

extruct/tool.py
lopuhin: And one more thing - since the public API of the tool.py changed, example in the README at the bottom also needs to be updated.

lopuhin:
@Kebniss @kmike this PR changes the public API of the main "extruct.extract" function (the way syntaxes are specified), so it's technically backwards incompatible. But I couldn't find any examples on github or in our internal code that uses syntaxes here, so it seems that this won't break anything we know. What do you think?

extruct/__init__.py
```diff
@@ -1,16 +1,38 @@
...
+from extruct.microformat import MicroformatExtractor
 from extruct.xmldom import XmlDomHTMLParser
 
 
-def extract(htmlstring, url='http://www.example.com/', encoding="UTF-8"):
+logger = logging.getLogger(__name__)
+
+
+def extract(htmlstring, url='http://www.example.com/', encoding="UTF-8",
+            syntaxes="all", schema_context='http://schema.org'):
```
kmike: what is schema_context argument for?

kmike:
> this PR changes the public API of the main "extruct.extract" function (the way syntaxes are specified)

Is it really changed? It was not possible to specify syntaxes before.

>  I'm ok with changing the API, but would like it to be more strict in terms of what it accepts - don't allow strings except for "all" and don't allow invalid format names.

+1. Also, it could be better to use None instead of "all", because it makes it less likely users will pass a string like "rdfa" instead of `["rdfa"]` list. Though it may be not important as soon as there is validation for this argument.

extruct/microformat.py
Kebniss: this page for example: https://shop.wwe.com/aj-styles-fastlane-2018-15-x-17-framed-plaque-w%2F-ring-canvas/CN1148.html?dwvar_CN1148_color=No%20Color#curr=USD&start=2
if you do 
```import requests
import mf2py

resp = requests.get(url, timeout=30)
mf2py.parse(resp.content, html_parser='lxml')['items'][10]```
This is not a product item but it's the structure it flattens. I could not find an example for a product with this structure but I guess they are out there. I imagine this was added to aid with making extracted content more uniform.

File: extruct/__init__.py (hunk 21 above)
Comment (Kebniss) [reply extruct/__init__.py]: it was needed by the tranformation part and forgot to take it away

File: extruct/microformat.py (hunk 14 above)
Comment (lopuhin) [reply extruct/microformat.py]: @Kebniss maybe the page changed, I get this:
```
{'children': [{'properties': {'name': ['Buy One, Get One for $5']},
   'type': ['h-hidden-phone']},
  {'properties': {'name': ['*Offer Ends Mar 28 02:59 AM EDT']},
   'type': ['h-hidden-phone']}],
 'properties': {'name': ['Buy One, Get One for $5\n-SHOP NOW-\n\n\n*Offer Ends Mar 28 02:59 AM EDT\nView Details\n\n\n\nBuy One Tee, Get One for $5. No code necessary. Exclusions apply (Preorder/Backorder, WrestleMania 34, Print on Demand, Homage, Finn Balor OGBC T-Shirt, Clearance). Cannot be combined with any other offer.\nEnds Mar 28\xa002:59 AM EDT']},
 'type': ['h-hidden-phone']}
```
which does not have ``'h-entry'`` type. 

> I imagine this was added to aid with making extracted content more uniform.

I'm mostly worried that in some cases this might loose valuable information.

File: extruct/__init__.py [hunk 22]
```diff
@@ -8,29 +8,29 @@
 from extruct.xmldom import XmlDomHTMLParser
 
 logger = logging.getLogger(__name__)
-VALID = ('microdata', 'opengraph', 'jsonld', 'microformat', 'rdfa')
+VALID = ('microdata', 'opengraph', 'json-ld', 'microformat', 'rdfa')
 
 def extract(htmlstring, url='http://www.example.com/', encoding="UTF-8",
-            syntaxes=['microdata', 'opengraph', 'jsonld', 'microformat', 'rdfa'],
+            syntaxes=['microdata', 'opengraph', 'json-ld', 'microformat', 'rdfa'],
```
lopuhin: can ``VALID`` constant be used here and below, and in tool.py, where we need a list of supported microformats?

extruct/microformat.py
Kebniss: That's correct, I just wanted to give you an example of the type of structure it flattens. I did not find any of these structures under an h-entry. 
We are losing valuable information because we are only returning `h-entry` objects. Microformat has entities specific for products that we should add. I mentioned this in issue #66.

lopuhin: > We are losing valuable information because we are only returning h-entry objects. 

Did you mean "we are *not* loosing"? If yes, will it change if we start returning other objects besides ``h-entry``?

extruct/__init__.py
```diff
@@ -1,16 +1,52 @@
...
+        processors.append(('microformat', MicroformatExtractor().extract_items))
+    if 'rdfa' in syntaxes:
+        processors.append(('rdfa', RDFaExtractor().extract_items))
+    output = {}
+    for label, extract in processors:
+        try:
+            output[label] = [obj for obj in extract(document=tree,
+                                                    url=url,
+                                                    html=htmlstring)]
+        except Exception:
```
lopuhin: is ``errors='strict'`` supported?

lopuhin:
@Kebniss I have only two questions left, https://github.com/scrapinghub/extruct/pull/61/files#r177656805 and https://github.com/scrapinghub/extruct/pull/61/files#r177655970, otherwise PR looks good to me.

extruct/microformat.py
Kebniss: My bad I read the code wrong. We are already returning all types of microformat and only flattening h-entry. It could make sense to simply return everything as is and then apply transformations in another library.

extruct/__init__.py
```diff
@@ -44,9 +44,11 @@ def extract(htmlstring, url='http://www.example.com/', encoding="UTF-8",
...
                                                     url=url,
                                                     html=htmlstring)]
-        except Exception:
+        except Exception as e:
             if errors == 'log':
                 logger.exception("Failed to parse %s", url)
             if errors == 'ignore':
                 pass
+            if errors == 'strict':
+                raise e
```
lopuhin: It's better to use ``raise`` instead of ``raise e`` in order to preserve original traceback

lopuhin:
@Kebniss I think this comment https://github.com/scrapinghub/extruct/pull/61#pullrequestreview-107786579 is still relevant, could you please have a look?

extruct/__init__.py
Kebniss: Ups, I missed this!

Issue #62, opened 2018-03-21
JSON Comment Removal RE may remove comments from within JSON strings


cathalgarvey:
(The regex used to remove comments)[https://github.com/scrapinghub/extruct/blob/c465e629c9e35cff08a703f6d2912c1c71c642ff/extruct/jsonld.py#L13] from JSON that fails to decode is pinned to the beginning on one side, but not the other. So, the regex may remove HTML comments from strings within the JSON document, as well as outside the JSON.

A quick fix would be to just add the `^` token to the other pattern, or to brace the two patterns to share the same `^`. But, I'm wondering whether trailing comments are also an occasional problem, and if so whether a custom little FSM scanner might be a better solution? E.g., something that scans for the earliest possible "valid" opening character for a JSON document, and the last such character, and returns the indices for those two characters for decoding?

PR #65, opened 2018-03-22
Fixing regex pattern to avoid removing comments from within valid JSON

cathalgarvey:
See #62 for premise. This fixes the issue, and adds test-cases. The test cases duplicate the schema.org invalid files; I didn't want to edit them in place, as it would confuse their provenance.

I still wonder whether regex is the wrong way to fix prefixed/suffixed garbage, but a "correct" solution might be less efficient and harder to maintain. So perhaps it's best to shelve "correct" for now. :)

### DIFF
--- a/extruct/jsonld.py
+++ b/extruct/jsonld.py
@@ -10,7 +10,7 @@ import lxml.etree
 import lxml.html
 
 
-HTML_OR_JS_COMMENTLINE = re.compile('^(\s*//.*)|(\s*<!--.*-->\s*)')
+HTML_OR_JS_COMMENTLINE = re.compile('^\s*(//.*|<!--.*-->)')
 
 class JsonLdExtractor(object):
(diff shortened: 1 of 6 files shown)


Kebniss:
Thank you for fixing this :)

kmike:
Thanks @cathalgarvey! The fix looks good to me. 

As for a "proper" solution, the problem is that data inside these tags is not JSON, it looks more like JavaScript, so likely it should be parsed with a JavaScript parser like https://github.com/rspivak/slimit, not with a JSON parser. 

But I recall these parsers are not always complete, and they can slow things down, so starting with just comment stripping using a regex sounds fine to me - unless there is a JS parser which is robust and fast enough.

There is an additional gotcha with a "proper" solution: HTML comments still have to be striped before passing to JS parser (see https://github.com/scrapy/parsel/issues/43).

PR #68, opened 2018-03-28
Modify raise in extract

Kebniss:


extruct/__init__.py
```diff
@@ -44,11 +44,12 @@ def extract(htmlstring, url='http://www.example.com/', encoding="UTF-8",
...
-        except Exception as e:
+        except Exception:
             if errors == 'log':
-                logger.exception("Failed to parse %s", url)
+                logger.exception("Failed to extract {} from {}".format(label, url))
             if errors == 'ignore':
                 pass
             if errors == 'strict':
-                raise e
+                print("Failed to extract {} from {}".format(label, url))
```
lopuhin: I think this ``print`` it not needed, and we already have errors="log": say you are a user of the library and want to know about an exception but then ignore it (or log yourself). Then you set errors="string", but still get this printed.

lopuhin:
thanks @Kebniss !

PR #71, opened 2018-04-03
Fix urljoin in microdata

Kebniss:
use w3lib.html.strip_html5_whitespace to strip whitespaces, newlines, etc from extracted urls and set default base url to None to avoid wrong results in urljoin


kmike:
Looks good 👍 
Could you please add a test case for this change? Check that url is joined correctly if there are whitespaces (I think there is no need to check all attributes), and that whitespaces are preserved for other fields.

lopuhin:
Looks good, thanks for fixing this @Kebniss!

>  set default base url to None to avoid wrong results in urljoin

example.com is also used as default url in individual extractors as well, I think it would be nice to change them too: https://github.com/scrapinghub/extruct/search?l=Python&q=example.com&type=

tests/samples/schema.org/product.json
```diff
@@ -0,0 +1,19 @@
+[{"type": "http://schema.org/Product",
+  "properties": {"brand": "ACME",
+                 "name": "Executive Anvil",
+                 "image": "anvil_executive.jpg",
+                 "description": "Sleeker than ACME's Classic Anvil, the\n      Executive Anvil is perfect for the business traveler\n      looking for something to drop from a height.",
```
kmike: It looks suspicious, as whitespace in the beginning and at the end seems to be stripped from the value. But this is not caused by this PR.

Kebniss:
How can I put this change in the CHANGELOG?

lopuhin:
@Kebniss the changelog is this file: https://github.com/scrapinghub/extruct/blob/master/HISTORY.rst - new features and changes are put there before (I think it's before, not after?) the release

Kebniss:
Should I add it now or only when we do the new release?

PR #76, opened 2018-05-03
0.5.0 release

Kebniss:


HISTORY.rst
```diff
@@ -2,6 +2,17 @@
...
 
+v0.5.0 (2018-05-03)
+-------------------
+
+* Add OpenGraph and Microformat.
+* Add argument ``syntaxes``, it allows to select which syntaxes to extract.
+* Add argument ``uniform``, if True it maps the output of Microdata, OpenGraph,
+  Microformat and Json-ld to the same template.
+* Add argument ``errors``, it allows to define if errors should be raised,
+  logged or ignored.
```
kmike: I think it makes sense to expand these notes: they say that arguments are added, but not explain for which function.

kmike:
Hey! I think not all changes are listed here, including backwards incompatible. For example,

1) https://github.com/scrapinghub/extruct/pull/71 - here default url is changed for all markup types, it is no longer `example.com`.  This is backwards incompatible, even though it is a bug fix.

2) Also, in https://github.com/scrapinghub/extruct/pull/71 extraction of microdata attributes is fixed. Microdata is not a new format, so we should mention this bug fix.

3) memory leak is fixed (https://github.com/scrapinghub/extruct/pull/60)

There are other pull requests merged since 0.4 release. Could you please go through them and update the changelog accordingly?

Kebniss:
> #71 - here default url is changed for all markup types, it is no longer example.com. This is backwards incompatible, even though it is a bug fix.

If it is backward incompatible then shouldn't the new release be the 1.0.0?

lopuhin:
> If it is backward incompatible then shouldn't the new release be the 1.0.0?

@Kebniss when major version is 0, this indicates that API is not stable yet and backwards incompatible changes are [possible](https://semver.org/#spec-item-4) between 0.x and 0.(x+1), although we try to keep them to the minimum

lopuhin:
I think https://github.com/scrapinghub/extruct/pull/71 is not fully covered in release notes - besides setting default url to None, it also strips whitespace. Apart from that, everything else seems to be covered.

Kebniss:
Codecov doesn't like me :(

HISTORY.rst
```diff
@@ -2,6 +2,46 @@
 History
 =======
 
+v0.5.0 (2018-05-09)
```
kmike: Date is no longer correct.

I think usually it makes sense to do a pull request for a changelog, and then, after it is merged, bump version and create a tag, using bumpversion.

kmike:
Thanks @Kebniss!

PR #78, opened 2018-05-14
Move code from __init__ module to _extruct

lopuhin:
The namespace would be cleaner if we keep only the imports in ``__init__``.
Not naming the new module ``extruct.extruct`` as this would interfere with
``extruct.extract`` autocompletion.

(another take instead of #77, but I still could not make git to show the diff for the move - only changes are several extra white spaces and an indentation fix).


Kebniss:
> I still could not make git to show the diff for the move - only changes are several extra white spaces and an indentation fix

What do you mean? Under ``Files changed`` I see the deleted lines in __init__ and the same added in _extruct

lopuhin:
> What do you mean? Under Files changed I see the deleted lines in init and the same added in _extruct

I would prefer the git to show the move and highlight only lines that are different between old ``__init__.py`` and new ``_extruct.py``

kmike:
👍 I like this style of import-only `__init__.py` a lot more (though that is not what Scrapy uses).
However, what do you think about keeping JsonLdExtractor and other classes at top level, to make it more backwards compatible?

lopuhin:
> However, what do you think about keeping JsonLdExtractor and other classes at top level, to make it more backwards compatible?

I didn't realize that, let me fix. Thanks for the catch!

kmike:
I wonder why codecov thinks coverage is decreased - it seems it thinks you're adding code without removing it, looks like a glitch.

lopuhin:
thanks for review @Kebniss and @kmike !

Kebniss:
Should this PR be logged in the changelog? It does not change anything significant in practice, correct?

lopuhin:
> It does not change anything significant in practice, correct?

yes, I think that's fair

PR #79, opened 2018-05-14
Use base_url instead of url

### LINKED ISSUE #75: url vs. base_url in extruct.extract
Right now ``extruct.extract`` has an ``url`` parameter which is documented as "url of the html documents", but in reality it's used as a base url (at least in ``LxmlMicrodataExtractor``, maybe in others as well). I think we should check if it's indeed always used as a base url, update documentation and introduce ``base_url`` argument deprecating ``url``? Another option would be to extract ``base_url`` in extruct, but this feels like worse solution to me (what if caller already has base_url or has more accurate base_url?), although we could also support both base_url and url.

lopuhin:
Fixes #75 

Changes:
- ``base_url`` is used instead of ``url`` in ``extruct.extract``, ``url`` is still supported by deprecated
- individual extractors accpet ``base_url`` instead of ``url``, unused keyword arguments are removed

TODO: 

- [x] add tests that make sure base_url is used
- [x] add tests for using deprectated url
- [x] rebase

### DIFF
--- a/extruct/microformat.py
+++ b/extruct/microformat.py
@@ -1,10 +1,11 @@
 import mf2py
 
+
 class MicroformatExtractor(object):
 
-    def extract(self, htmlstring, url=None, encoding='UTF-8'):
-        return list(self.extract_items(htmlstring, url=url))
+    def extract(self, htmlstring, base_url=None, encoding='UTF-8'):
+        return list(self.extract_items(htmlstring, base_url=base_url))
 
-    def extract_items(self, html, url, document=None):
-        for obj in mf2py.parse(html, html_parser="lxml", url=url)['items']:
+    def extract_items(self, html, base_url=None):
+        for obj in mf2py.parse(html, html_parser="lxml", url=base_url)['items']:
             yield obj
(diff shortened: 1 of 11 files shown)


kmike:
> In the changelog should I mention all the functions that are made private in this PR? Eg in _extract-item() w3microdata

yes 👍

extruct/microformat.py
```diff
@@ -1,10 +1,11 @@
...
 
+
 class MicroformatExtractor(object):
 
-    def extract(self, htmlstring, url=None, encoding='UTF-8'):
-        return list(self.extract_items(htmlstring, url=url))
+    def extract(self, htmlstring, base_url=None, encoding='UTF-8'):
+        return list(self.extract_items(htmlstring, base_url=base_url))
 
-    def extract_items(self, html, url, document=None):
```
kmike: 👍

kmike:
Changing argument name from url to base_url in JsonLdExtractor, etc. is backwards incompatible (unlike extruct.extract, where there is a fallback). I'm not sure we should provide a fallback everywhere, but it should be mentioned in the release notes.

PR #80, opened 2018-06-04
Fix tests for microformats

jakubwasikowski:
It seems that new version of the library for parsing microformats have been updated recently and it causes that three tests does not pass anymore. I did the following here:
- I verified why exactly the tests fails. 
  * First of all `mf2py` seems to trim text in the content, so it means that signs of new line are removed. 
  * Second thing is that in some cases the name property has been removed from output. To be honest I don't understand why it happened, especially that it is not removed every time. Do you think we should care about that and ask them about that?
- I believe that we can just use this new version so I fixed all tests.
- I fixed the version of library in `requirements.txt` to 1.1.0.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,6 @@ gevent
 requests
 rdflib
 rdflib-jsonld
-mf2py==1.1.0
+mf2py>=1.1.0
 six
 w3lib
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,6 @@ gevent
 requests
 rdflib
 rdflib-jsonld
-mf2py
+mf2py==1.1.0
 six
 w3lib
--- a/tests/samples/misc/microformat_flat_test.json
+++ b/tests/samples/misc/microformat_flat_test.json
@@ -1,78 +1,75 @@
 [
-  {
-    "@type": [
-      "h-hidden-tablet",
-      "h-hidden-phone"
-    ],
-    "@context": "http://microformats.org/wiki/",
-    "name": [
-      ""
-    ]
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -6,6 +6,6 @@ gevent
 requests
 rdflib
 rdflib-jsonld
-mf2py
+mf2py==1.1.0
```
kmike: I think it is good to run tests on latest mf2py version, not only on 1.1.0. Maybe change it to >=?

jakubwasikowski:
@Kebniss, you can notice that in tests/test_uniform.py where I had to remove `'name': ['']` property because it doesn't appear in the result anymore. And you can also see there that not all empty properties like `'name': ['']` have been removed from the output - for instance there is still one in the first item of the `expected` array. I checked the html and I didn't notice any special difference between these cases, that for the first the property is removed and for the second not.

requirements.txt
jakubwasikowski: Actually, `1.1.0` is the latest version of `mf2py`

Kebniss:
I see. if it happens only with empty strings it is not be a problem then :)

requirements.txt
kmike: Yes, but if there will be a new version, our tests won't be able to check how it works, we'd have to track these versions manually, updating it from time to time.

jakubwasikowski: Sorry, I didn't get your first comment - now I understand :) Sure, will do that :+1:

kmike:
Thanks @jakubwasikowski!

PR #81, opened 2018-06-04
[MRG+1] Add option to return html node during extraction

jakubwasikowski:
I've added here an option to return html nodes in the result of `extract` method. The motivation of that change is to be able to parse additionally html node in case the metadata item is not fully consistent with the standard.
I checked in which syntaxes I can add that option and it seems it's possible only for microdata. 

I did here the following:
- Added returning node in microdata parser.
- Added option to `extract` method for returning html node.
- Updated tests. Please take a look in tests, because I had to do replacing of node reference with the identifier of node in order to have test case in json. Let me please know if you do not like it. 

TODO:
- [x]  add a note about a key under which the node is returned;
- [x] mention that node is lxml.etree.Element (or what type is the node?);
- [x] wrap to 79 columns;
- [x] remove unneeded blank lines, leaving 1?
- [x] add test for uniform=True
- [x] document this new feature in README
- [x] wrap to 79 columns in `history.rst`
- [x] specify in `history.rst` that new option is available only for microdata and which argument allows you to use that.
- [x] move info about supporting to the beginning
- [x] add information about supporting to the comment of `extract`

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -31,7 +31,7 @@ v0.5.0 (TBD)
 * In ``w3microdata`` ``_extract_properties``, ``_extract_property_refs``, 
   ``_extract_property``, ``_extract_property_value`` and ``_extract_item``
   now need ``items_seen`` and ``url`` to be passed as arguments.
-* Add option in ``extruct.extract`` to return HTML node with the return of metadata extraction.
+* Add the option in ``extruct.extract`` to return HTML node with the result of metadata extraction.
 
 Warning: backward-incompatible change:
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -31,6 +31,7 @@ v0.5.0 (TBD)
 * In ``w3microdata`` ``_extract_properties``, ``_extract_property_refs``, 
   ``_extract_property``, ``_extract_property_value`` and ``_extract_item``
   now need ``items_seen`` and ``url`` to be passed as arguments.
+* Add option in ``extruct.extract`` to return HTML node with the return of metadata extraction.
 
 Warning: backward-incompatible change:
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -32,8 +32,8 @@ v0.5.0 (TBD)
   ``_extract_property``, ``_extract_property_value`` and ``_extract_item``
   now need ``items_seen`` and ``url`` to be passed as arguments.
 * Add argument ``return_html_node`` to ``extract``, it allows to return HTML
-node with the result of metadata extraction. It is supported only by microdata
-syntax.
+  node with the result of metadata extraction. It is supported only by
+  microdata syntax.
 
 Warning: backward-incompatible change:
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -31,7 +31,9 @@ v0.5.0 (TBD)
 * In ``w3microdata`` ``_extract_properties``, ``_extract_property_refs``, 
   ``_extract_property``, ``_extract_property_value`` and ``_extract_item``
   now need ``items_seen`` and ``url`` to be passed as arguments.
-* Add the option in ``extruct.extract`` to return HTML node with the result of metadata extraction.
+* Add argument ``return_html_node`` to ``extract``, it allows to return HTML
+node with the result of metadata extraction. It is supported only by microdata
+syntax.
 
 Warning: backward-incompatible change:
 
--- a/README.rst
+++ b/README.rst
@@ -267,7 +267,6 @@ As the result, an additional key `nodeHtml` will be included in the result for e
 item. The node is of `lxml.etree.Element` type.
 The feature is supported only by syntaxes that are based on HTML.
 
-::
   >>> r = requests.get('http://www.rugpadcorner.com/shop/no-muv/')
   >>> base_url = get_base_url(r.text, r.url)
   >>> data = extruct.extract(r.text, base_url, syntaxes=['microdata'], return_html_node=True)
--- a/README.rst
+++ b/README.rst
@@ -262,9 +262,9 @@ Returning HTML node
 +++++++++++++++++++
 
 It is also possible to get references to HTML node for every extracted metadata item.
-To make it happen, just set the `return_html_node` option of `extract` method to `True`.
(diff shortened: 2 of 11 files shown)


jakubwasikowski:
Checks should pass when we merge https://github.com/scrapinghub/extruct/pull/80.

extruct/_extruct.py
```diff
@@ -34,6 +35,7 @@ def extract(htmlstring, base_url=None, encoding="UTF-8",
                  '@type': 'example_type',
                  /* All other the properties in keys here */
                  }
+       return_html_node: if True, it includes HTML node of respective embedded metadata into the result.
```
kmike: I think it makes sense to 

* and add a note about a key under which the node is returned;
* mention that node is lxml.etree.Element (or what type is the node?);
* wrap it to 79 columns;

tests/test_extruct.py
```diff
@@ -17,25 +17,50 @@ def test_all(self):
...
+                if key == "htmlNode":
+                    item["_nodeId_"] = val.get("id")
+                    del item[key]
+                else:
+                    self._replace_node_ref_with_node_id(val)
+
+
+
+
+
```
kmike: could you please remove unneeded blank lines, leaving 1?

tests/test_extruct.py
```diff
@@ -17,25 +17,50 @@ def test_all(self):
...
-        body, expected = self._microdata_custom_url()
+        body, expected = self._microdata_custom_url('product_custom_url.json')
         data = extruct.extract(body, base_url='http://some-example.com',
                                syntaxes=['microdata'])
         self.assertEqual(data, expected)
 
+    def test_microdata_with_returning_node(self):
+        body, expected = self._microdata_custom_url('product_custom_url_and_node_id.json')
+        data = extruct.extract(body, base_url='http://some-example.com',
+                               syntaxes=['microdata'], return_html_node=True)
```
kmike: I think there should also be a test with uniform=True

kmike:
Hey @jakubwasikowski! The fix looks good; I've left a few small comments. Could you please also update README, so that this feature is documented?

extruct/_extruct.py
jakubwasikowski: Sure, will do that :+1:

README.rst
```diff
@@ -258,6 +258,41 @@ To do so set ``uniform=True`` when calling ``extract``, it's false by default fo
...
 NB rdfa structure is not uniformed yet
 
+Returning HTML node
++++++++++++++++++++
+
+It is also possible to get references to HTML node for every extracted metadata item.
+To make it happen, just set the ``return_html_node`` option of ``extract`` method to ``True``.
+As the result, an additional key "nodeHtml" will be included in the result for every
+item. The node is of ``lxml.etree.Element`` type.
+The feature is supported only by syntaxes that are based on HTML: ::
```
kmike: microformats are also embedded in html, but we don't support returning nodes for them, as it is not implemented. I think it makes more sense to say which syntaxes are supported, explicitly. One may argue that OpenGraph is also based on HTML (annotations are in html meta tags), or even json-ld (annotations are in html, inside a special tag).

jakubwasikowski: done :+1:

HISTORY.rst
```diff
@@ -31,6 +31,7 @@ v0.5.0 (TBD)
 * In ``w3microdata`` ``_extract_properties``, ``_extract_property_refs``, 
   ``_extract_property``, ``_extract_property_value`` and ``_extract_item``
   now need ``items_seen`` and ``url`` to be passed as arguments.
+* Add the option in ``extruct.extract`` to return HTML node with the result of metadata extraction.
```
Kebniss: Can you split this line so it is no longer than 79 columns? :)

README.rst
```diff
@@ -258,6 +258,41 @@ To do so set ``uniform=True`` when calling ``extract``, it's false by default fo
 
 NB rdfa structure is not uniformed yet
 
+Returning HTML node
++++++++++++++++++++
+
+It is also possible to get references to HTML node for every extracted metadata item.
```
Kebniss: You should specify that this option is available only for microdata. If a user selects return_html_node = True and syntaxes = [opengraph] it won't see the nodes even though it will expect to

extruct/_extruct.py
```diff
@@ -34,6 +35,9 @@ def extract(htmlstring, base_url=None, encoding="UTF-8",
                  '@type': 'example_type',
                  /* All other the properties in keys here */
                  }
+       return_html_node: if True, it includes into the result a HTML node of
+                         respective embedded metadata under 'htmlNode' key.
```
Kebniss: I would add that is only microdata compatible here as well. Some people might just read the comments and miss this piece of information from the docs :)

Kebniss:
There are just a couple of minor things in the comments but apart from that the PR is good :)
Thank you for adding this option @jakubwasikowski!

HISTORY.rst
jakubwasikowski: Agree, I'll fix both.
By the way concerning these 79 columns, do we follow this rule in the whole project? I noticed that there are also some other lines in both code and text files that are longer than 79 columns. I know that these 79 columns are specified by PEP8 standard but my personal opinion about that is that 120 characters is more optimal and 79 characters are too less for current screen resolutions.
In addition, I can see in PEP8 that the rules are even more restricted and "for flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.".
What do you guys think? What rules do we accept?

kmike: @jakubwasikowski we're not strict about it, but I think a convention which we follow is to use 79 columns - both for code and docstrings, unless there is a reason not to do so. 

Regarding "reason": introducing a local variable to make an expression fit to a limit is fine; splitting a string constant with a long URL just to fit it to 79 columns is not a good idea.

It may also depend on a project: it is fine to use 120 columns limit if you notice that the project doesn't respect 79 columns limit often; if code is mostly fit into 79 columns, it is better to go with that limit, so that the code is consistent - unless there is a strong reason to diverge.

jakubwasikowski:
@Kebniss, the code is ready for re-review.

Kebniss:
PR approved, thank you again @jakubwasikowski!

Issue #31, opened 2016-12-15
Support "expanded" Open Graph metadata based on og:type


redapple:
Facebook Oepn Graph defines an [expanded version of embedded metadata](http://ogp.me/#types) depending on the value of `og:type`.

For example:

```
article - Namespace URI: http://ogp.me/ns/article#

    article:published_time - datetime - When the article was first published.
    article:modified_time - datetime - When the article was last changed.
    article:expiration_time - datetime - When the article is out of date after.
    article:author - profile array - Writers of the article.
    article:section - string - A high-level section name. E.g. Technology
    article:tag - string array - Tag words associated with this article.
```

This is used for example on nytimes.com. Snippet:

```
<meta property="og:url" content="http://www.nytimes.com/2016/12/15/arts/music/from-steet-theater-to-wagner-on-the-opera-stage.html" />
<meta property="og:type" content="article" />
<meta property="og:title" content="From Street Theater to Wagner on the Opera Stage" />
<meta property="og:description" content="Àlex Ollé brings an avant-garde sensibility to “The Flying Dutchman,” which he set in Bangladesh instead of Norway. The production opens in Madrid on Saturday." />
<meta property="article:published" itemprop="datePublished" content="2016-12-15T05:55:55-05:00" />
<meta property="article:modified" itemprop="dateModified" content="2016-12-15T06:19:30-05:00" />
<meta property="article:section" itemprop="articleSection" content="Music" />
<meta property="article:section-taxonomy-id" itemprop="articleSection" content="C5BFA7D5-359C-427B-90E6-6B7245A6CDD8" />
<meta property="article:section_url" content="http://www.nytimes.com/section/arts" />
<meta property="article:top-level-section" content="arts" />
<meta property="fb:app_id" content="9869919170" />
```

Currently (as I write these lines version 0.3.0a1) extracts raw `article:...` properties 

```
...
  'article:author': [{'@value': 'http://www.nytimes.com/by/raphael-minder'}],
  'article:collection': [{'@value': 'https://static01.nyt.com/services/json/sectionfronts/arts/music/index.jsonp'}],
  'article:modified': [{'@value': '2016-12-15T06:19:30-05:00'}],
  'article:published': [{'@value': '2016-12-15T05:55:55-05:00'}],
  'article:section': [{'@value': 'Music'}],
  'article:section-taxonomy-id': [{'@value': 'C5BFA7D5-359C-427B-90E6-6B7245A6CDD8'}],
[...]

PR #88, opened 2018-08-21
Handle known opengraph namespaces

jakubwasikowski:
Added the following things for OpenGraph:
- Defined and handled known OpenGraph namespaces.
- Added compiling regex.
- Add extraction of prefixes from html element

### DIFF
--- a/extruct/opengraph.py
+++ b/extruct/opengraph.py
@@ -13,6 +13,10 @@ _OG_NAMESPACES = {
 }
 
 
+def _merge_dicts(dict1, dict2):
+    return {**dict1, **dict2}
+
+
 class OpenGraphExtractor(object):
     """OpenGraph extractor following extruct API."""
 
@@ -23,16 +27,17 @@ class OpenGraphExtractor(object):
 
     def extract_items(self, document, base_url=None):
         # OpenGraph defines a web page as a single rich object.
-        # TODO: Handle known opengraph namespaces.
         for head in document.xpath('//head'):
-            prefix = dict(_PREFIX_PATTERN.findall(head.attrib.get('prefix', '')))
-            prefix.setdefault('og', 'http://ogp.me/ns#')
+            head_namespaces = dict(
+                _PREFIX_PATTERN.findall(head.attrib.get('prefix', ''))
+            )
+            namespaces = _merge_dicts(_OG_NAMESPACES, head_namespaces)
             props = []
             for el in head.xpath('meta[@property and @content]'):
                 prop = el.attrib['property']
                 val = el.attrib['content']
                 ns = prop.partition(':')[0]
-                if ns in prefix:
+                if ns in namespaces:
                     props.append((prop, val))
             if props:
-                yield {'namespace': prefix, 'properties': props}
+                yield {'namespace': namespaces, 'properties': props}
--- a/extruct/opengraph.py
+++ b/extruct/opengraph.py
@@ -13,10 +13,6 @@ _OG_NAMESPACES = {
 }
 
 
-def _merge_dicts(dict1, dict2):
-    return {**dict1, **dict2}
-
-
 class OpenGraphExtractor(object):
     """OpenGraph extractor following extruct API."""
 
@@ -29,9 +25,9 @@ class OpenGraphExtractor(object):
         # OpenGraph defines a web page as a single rich object.
         for head in document.xpath('//head'):
             html_elems = document.head.xpath("parent::html")
-            html_nms = self.get_namespaces(html_elems[0]) if html_elems else {}
-            head_nms = self.get_namespaces(head)
-            namespaces = _merge_dicts(html_nms, head_nms)
+            namespaces = self.get_namespaces(
+                html_elems[0]) if html_elems else {}
+            namespaces.update(self.get_namespaces(head))
             props = []
             for el in head.xpath('meta[@property and @content]'):
                 prop = el.attrib['property']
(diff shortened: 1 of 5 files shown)


extruct/opengraph.py
```diff
@@ -2,6 +2,17 @@
 import lxml.html
 
 
+_PREFIX_PATTERN = re.compile(r'\s*(\w+): ([^\s]+)')
```
kmike: I wonder if we should be less strict regarding whitespaces after `:`; in http://ogp.me/ I found nothing about a single whitespace being required. Something like `r'\s*(\w+):\s*([^\s]+)'`?

extruct/opengraph.py
```diff
@@ -12,16 +23,24 @@ def extract(self, htmlstring, base_url=None, encoding='UTF-8'):
 
     def extract_items(self, document, base_url=None):
         # OpenGraph defines a web page as a single rich object.
-        # TODO: Handle known opengraph namespaces.
         for head in document.xpath('//head'):
-            prefix = dict(re.findall(r'\s*(\w+): ([^\s]+)', head.attrib.get('prefix', '')))
-            prefix.setdefault('og', 'http://ogp.me/ns#')
+            html_elems = document.head.xpath("parent::html")
```
kmike: 👍

kmike:
Looks great, thanks @jakubwasikowski! 

There is a "Roadmap" section in README - do you think it can be removed now?

I'm not sure I've read https://github.com/scrapinghub/extruct/issues/31 correctly, but it could also be related.

tests/samples/misc/opengraph_flat_test.json
```diff
@@ -5,9 +5,14 @@
       "og:image": "https://www.eventeducation.com/images/982336_wedding_dayandouan_th.jpg",
       "og:site_name": "Event Education",
       "og:description": "Event Education provides free courses on event planning and management to event professionals worldwide.",
+      "article:publisher": "http://www.facebook.com/PUBLISHER",
```
lopuhin: I wonder what was result of parsing ``<meta property="article:publisher" content="http://www.facebook.com/PUBLISHER" />`` before this PR?

extruct/opengraph.py
jakubwasikowski: I've tried to find if there is any specification about that, but didn't find any. Agree, will be better to put `\s*` there :+1:

tests/samples/misc/opengraph_flat_test.json
jakubwasikowski: Here is the result of parsing before that PR:
```
{
    "@context": {
        "og": "http://ogp.me/ns#"
    },
    "@type": "article",
    "og:description": "Event Education provides free courses on event planning and management to event professionals worldwide.",
    "og:image": "https://www.eventeducation.com/images/982336_wedding_dayandouan_th.jpg",
    "og:site_name": "Event Education",
    "og:title": "Himanshu's Open Graph Protocol",
    "og:url": "https://www.eventeducation.com/test.php"
}
```
And after:
```
{
    "@context": {
        "article": "http://ogp.me/ns/article#",
        "og": "http://ogp.me/ns#"
    },
    "@type": "article",
    "article:author": "http://facebook.com/AUTHOR",
    "article:modified_time": "2016-12-13T12:35:52-07:00",
    "article:published_time": "2012-04-04T19:50:00-07:00",
    "article:publisher": "http://www.facebook.com/PUBLISHER",
    "og:description": "Event Education provides free courses on event planning and management to event professionals worldwide.",
    "og:image": "https://www.eventeducation.com/images/982336_wedding_dayandouan_th.jpg",
    "og:site_name": "Event Education",
    "og:title": "Himanshu's Open Graph Protocol",
    "og:url": "https://www.eventeducation.com/test.php"
}
```
Property `article:publisher` would be in the output for the version before only if there is prefix for articles defined in `<head>` tag, but for that test case it doesn't take place. Furthermore I noticed in our test case that this is quite rare situation.

jakubwasikowski:
> I'm not sure I've read #31 correctly, but it could also be related.

@kmike, I'm not sure what @redapple exactly meant by "while they could use the type-dependent OGP namespace", but I think this is suggestion that to implement interpretation of value of property (for instance casting to specified type) depending on the type of schema from `og:type`.

I'm not doing that in the PR - I've just defined that some schemas are known and "pre-defined" in OGP and there is no need to specify these schemas in `prefix` attribute. I'm not interpreting here values of these properties.

tests/samples/misc/opengraph_flat_test.json
lopuhin: @jakubwasikowski thanks for an example, I didn't realize we were missing fields from unknown schemas here 👍

jakubwasikowski:
> There is a "Roadmap" section in README - do you think it can be removed now?

I think it can be removed and I did that, but there are two things:
- does that entry in roadmap relate to interpreting property type depending on type of schema?
- I can see that on the page mentioned in README (https://developers.facebook.com/docs/sharing/opengraph/object-properties) there are more schemas than these specified in http://ogp.me/#types, but I have to say that there is a lot of mess and it's quite hard to say what schemas should be defined. Despite that fact, do we want to support some of them?

If there is `yes` answer to any of above questions then we should rather keep that roadmap section.

extruct/opengraph.py
jakubwasikowski: Done

kmike:
@jakubwasikowski good points. Though I think either way README is a wrong place for a roadmap, and supporting all this stuff was not in *our* roadmap anyways, it seems priorities changed.

kmike:
Thanks Jakub!

PR #97, opened 2018-11-01
better Python 3.7 support

kmike:
* added Python 3.7 to tests
* fix Python 3.7 warnings
* improve DeprecationWarning, so that it points to a caller code

### DIFF
--- a/extruct/tool.py
+++ b/extruct/tool.py
@@ -14,8 +14,8 @@ def metadata_from_url(url, syntaxes=SYNTAXES, uniform=False,
         resp.raise_for_status()
     except requests.exceptions.HTTPError:
         return result
-    result.update(extruct.extract(resp.content, 
-                                  url=url, 
+    result.update(extruct.extract(resp.content,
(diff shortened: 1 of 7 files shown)


extruct/tool.py
```diff
@@ -14,8 +14,8 @@ def metadata_from_url(url, syntaxes=SYNTAXES, uniform=False,
         resp.raise_for_status()
     except requests.exceptions.HTTPError:
         return result
-    result.update(extruct.extract(resp.content, 
-                                  url=url, 
+    result.update(extruct.extract(resp.content,
+                                  base_url=url,  # FIXME: use base url
```
lopuhin: One more reason to deprecate the tool - I was going to suggest using get_base_url from w3lib, but it seems this would require more extra code to handle encodings correctly (and we are already throwing encoding information away here by not using headers).

kmike: Yeah, actually I started adding get_base_url, then realized encodings need to be handled, etc., removed that, and just added FIXME comment instead :)

Issue #99, opened 2018-11-28
errors="ignore" does not ignore all errors


lopuhin:
E.g. if an error happens during html parsing or unification, it won't be ignored when calling ``extrcuct.extract(html, errors='ignore')``

lopuhin:
Fixed by #100 , thanks @Kebniss !

PR #100, opened 2018-11-29
ignore any exception if errors='ignore'

Kebniss:
as explained in #99 errors='ignore' was not universal. 

I wanted to add a test but I don't know how to test exceptions

### DIFF
--- a/extruct/_extruct.py
+++ b/extruct/_extruct.py
@@ -61,7 +61,7 @@ def extract(htmlstring,
             return {}
         if errors == 'log':
             logger.exception(
-                f'Failed to parse html, raises {e}')
+                'Failed to parse html, raises {}'.format(e))
             return {}
         if errors == 'strict':
             raise e
@@ -101,7 +101,9 @@ def extract(htmlstring,
             output[syntax] = list(extract(document, base_url=base_url))
         except Exception as e:
             if errors == 'log':
-                logger.exception(f'Failed to extract {syntax}, raises {e}')
+                logger.exception('Failed to extract {}, raises {}'
+                                 .format(syntax, e)
+                                 )
             if errors == 'ignore':
                 pass
             if errors == 'strict':
@@ -141,7 +143,9 @@ def extract(htmlstring,
                 if errors == 'log':
                     output[syntax] = []
                     logger.exception(
-                        f'Failed to uniform extracted for {syntax}, raises {e}')
+                        'Failed to uniform extracted for {}, raises {}'
+                        .format(syntax, e)
+                        )
                 if errors == 'strict':
                     raise
 
--- a/extruct/_extruct.py
+++ b/extruct/_extruct.py
@@ -81,37 +81,58 @@ def extract(htmlstring,
             ('microformat', MicroformatExtractor().extract_items,
              htmlstring))
     if 'rdfa' in syntaxes:
-        processors.append(('rdfa', RDFaExtractor().extract_items, tree))
+        processors.append(
+            ('rdfa', RDFaExtractor().extract_items,
+            tree,
+            ))
     output = {}
-    for label, extract, document in processors:
+    for syntax, extract, document in processors:
         try:
-            output[label] = list(extract(document, base_url=base_url))
+            output[syntax] = list(extract(document, base_url=base_url))
         except Exception:
             if errors == 'log':
-                logger.exception('Failed to extract {}'.format(label))
+                logger.exception('Failed to extract {}'.format(syntax))
             if errors == 'ignore':
                 pass
             if errors == 'strict':
                 raise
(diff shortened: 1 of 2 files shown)


extruct/_extruct.py
```diff
@@ -38,48 +40,63 @@ def extract(htmlstring, base_url=None, encoding="UTF-8",
...
+                    output['microdata'], schema_context=schema_context)
+            if 'microformat' in syntaxes:
+                output['microformat'] = _umicrodata_microformat(
+                    output['microformat'],
+                    schema_context='http://microformats.org/wiki/')
+            if 'opengraph' in syntaxes:
+                output['opengraph'] = _uopengraph(output['opengraph'])
+    except Exception as e:
+        if errors == 'ignore':
+            return {}
```
lopuhin: I think ``errors == 'log'`` also needs to be handled here

extruct/_extruct.py
```diff
@@ -38,48 +40,63 @@ def extract(htmlstring, base_url=None, encoding="UTF-8",
...
+            if 'microdata' in syntaxes:
+                output['microdata'] = _umicrodata_microformat(
+                    output['microdata'], schema_context=schema_context)
+            if 'microformat' in syntaxes:
+                output['microformat'] = _umicrodata_microformat(
+                    output['microformat'],
+                    schema_context='http://microformats.org/wiki/')
+            if 'opengraph' in syntaxes:
+                output['opengraph'] = _uopengraph(output['opengraph'])
+    except Exception as e:
```
Kebniss: If uniform fails for only one syntax shouldn't we just set that to None and return the others? This would make the code a bit cluttered since I have to wrap every case in a try, except but I feel it makes more sense

lopuhin: Yes, that would be better, also I think it's possible to avoid repeating exception handling code for each unifyer in a similar way to how syntaxes are handled with a list of ``processors``. But also I think it's fine to not do it in this PR and treat as a possible future improvement.

extruct/_extruct.py
```diff
@@ -38,48 +40,109 @@ def extract(htmlstring, base_url=None, encoding="UTF-8",
...
+            uniform_processors.append(
+                ('opengraph',
+                 _uopengraph,
+                 output['opengraph'],
+                 None,
+                 ))
+        for syntax, uniform, raw, schema_context in uniform_processors:
+            try:
+                if syntax == 'opengraph':
+                        output[syntax] = uniform(raw)
```
lopuhin: Double indent here

extruct/_extruct.py
```diff
@@ -38,48 +40,109 @@ def extract(htmlstring, base_url=None, encoding="UTF-8",
...
+                    output[syntax] = uniform(raw, schema_context)
+            except Exception as e:
+                if errors == 'ignore':
+                    output[syntax] = []
+                if errors == 'log':
+                    output[syntax] = []
+                    logger.exception(
+                        f'Failed to uniform extracted, exception raised {e}')
+                if errors == 'strict':
+                    raise e
```
lopuhin: I think it's better to use just ``raise`` to save the original traceback
```suggestion
                    raise
```

extruct/_extruct.py
```diff
@@ -38,48 +40,109 @@ def extract(htmlstring, base_url=None, encoding="UTF-8",
...
-    for label, extract, document in processors:
+    for syntax, extract, document in processors:
         try:
-            output[label] = list(extract(document, base_url=base_url))
-        except Exception:
+            output[syntax] = list(extract(document, base_url=base_url))
+        except Exception as e:
             if errors == 'log':
-                logger.exception('Failed to extract {}'.format(label))
+                logger.exception(f'Failed to extract {syntax}, raises {e}')
```
lopuhin: f syntax appeared only in python 3.6, we'll have to use an older syntax for python 2.7+ 😢

Kebniss: ah, darn! It makes the code look much neater!

Kebniss:
codecov fails because some exceptions are not tested, however I cannot come up with an input that fails when uniforming data

extruct/_extruct.py
```diff
@@ -38,48 +40,113 @@ def extract(htmlstring, base_url=None, encoding="UTF-8",
...
+        tree = parse_xmldom_html(htmlstring, encoding=encoding)
+    except Exception as e:
+        if errors == 'ignore':
+            return {}
+        if errors == 'log':
+            logger.exception(
+                'Failed to parse html, raises {}'.format(e))
+            return {}
+        if errors == 'strict':
+            raise e
```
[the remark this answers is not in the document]
Kebniss: sorry, I thought I fixed all of these

PR #103, opened 2019-01-25
Add raw json output to JsonLdExtractor

Granitosaurus:
Sometimes jsonld schema is prefered in raw json format rather than python dict - this PR implementes `as_json` kwarg bool to determine whether to return python dictionary or json string.

Some use cases:
    
    1. Use alternative json library
    2. Use json.loads kwargs
    3. Use object hook functions

My personal use case is to strip away `@` keys with object hook:

    def strip_jsonld(data):
        return {k: v for k, v in data.items() if not k.startswith('@')}

    data = json.loads(script, object_hook=strip_jsonld)

### DIFF
--- a/extruct/jsonld.py
+++ b/extruct/jsonld.py
@@ -16,26 +16,24 @@ HTML_OR_JS_COMMENTLINE = re.compile(r'^\s*(//.*|<!--.*-->)')
 class JsonLdExtractor(object):
     _xp_jsonld = lxml.etree.XPath('descendant-or-self::script[@type="application/ld+json"]')
 
-    def extract(self, htmlstring, base_url=None, encoding="UTF-8"):
+    def extract(self, htmlstring, base_url=None, encoding="UTF-8", as_json=False):
         tree = parse_html(htmlstring, encoding=encoding)
-        return self.extract_items(tree, base_url=base_url)
+        return self.extract_items(tree, base_url=base_url, as_json=as_json)
 
-    def extract_items(self, document, base_url=None):
+    def extract_items(self, document, base_url=None, as_json=False):
         return [
             item
-            for items in map(self._extract_items, self._xp_jsonld(document))
+            for items in map(self._extract_items_raw if as_json else self.extract_items, self._xp_jsonld(document))
             if items for item in items if item
         ]
 
+    def _extract_items_raw(self, node):
+        return HTML_OR_JS_COMMENTLINE.sub('', node.xpath('string()'))
+
     def _extract_items(self, node):
-        script = node.xpath('string()')
-        try:
-            # TODO: `strict=False` can be configurable if needed
-            data = json.loads(script, strict=False)
-        except ValueError:
-            # sometimes JSON-decoding errors are due to leading HTML or JavaScript comments
-            data = json.loads(
-                HTML_OR_JS_COMMENTLINE.sub('', script), strict=False)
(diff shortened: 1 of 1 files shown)


extruct/jsonld.py
```diff
@@ -16,26 +16,24 @@
...
-        try:
-            # TODO: `strict=False` can be configurable if needed
-            data = json.loads(script, strict=False)
-        except ValueError:
-            # sometimes JSON-decoding errors are due to leading HTML or JavaScript comments
-            data = json.loads(
-                HTML_OR_JS_COMMENTLINE.sub('', script), strict=False)
+        script = self._extract_items_raw(node)
+        # TODO: `strict=False` can be configurable if needed
+        data = json.loads(script, strict=False)
```
lopuhin: Why is try/except removed here?

extruct/jsonld.py
```diff
@@ -16,26 +16,24 @@
 class JsonLdExtractor(object):
     _xp_jsonld = lxml.etree.XPath('descendant-or-self::script[@type="application/ld+json"]')
 
-    def extract(self, htmlstring, base_url=None, encoding="UTF-8"):
+    def extract(self, htmlstring, base_url=None, encoding="UTF-8", as_json=False):
```
lopuhin: To me ``as_json`` sounds confusing - it's not clear what will this return, parsed JSON or not parsed (I'd lean towards parsed, although in reality it's not parsed). I'm not sure which name would be better, maybe ``parse_json=True``?

Granitosaurus: I agree parse_json sounds nicer!

extruct/jsonld.py
Granitosaurus: Original code tries to load json and then remove comments. I can't imagine scenario where someone would like to receive json with javascript/html comments in tackt. I took the liberty of simplifying the process and explicitly remove the comments as it's relatively cheap operation. If you disagree we can change it back to "trying to remove comments if loading fails" approach.

lopuhin: @Granitosaurus ah I didn't notice that it's still there but was moved, thanks for pointing it out. I'd rather not try removing comments by default, because they can be in valid JSON:
```
>>> json.dumps(['<!-- comment -->'])
'["<!-- comment -->"]'
```
so removing comments it a heuristics which is not always correct. So I think it would be better to do just `node.xpath('string()'))` in `_extract_items_raw`, and try removing comments only if json parsing fails, as in master.

Granitosaurus: Sure thing, reverted!

Granitosaurus:
Also added this feature to RDFA extractor and updated readme with short examples.

Gallaecio:
This looks mostly ready. What about adding a test covering the usage of `parse_json=False`?

Granitosaurus:
Sorry that it took me so long to attend to this but the tests gave me a bit of an headache :D 

Should be good to go now!

Gallaecio:
`TestRDFa.test_wikipedia_xhtml_rdfa_no_prefix` seems to be failing after your changes.

Granitosaurus:
> `TestRDFa.test_wikipedia_xhtml_rdfa_no_prefix` seems to be failing after your changes.

It has been failing master for me too; I actually update the fixtures in this PR to prevent it failing but the test just seems to be flawed in some sense - it keeps either breakin on travis or on my machine locally. :man_shrugging:

PR #105, opened 2019-02-06
Avoid including itemprop from child itemscopes when using itemref

ivanprado:
Extruct deals wrongly with elements referenced by `itemref` if the referenced element also include other items defined inside with `itemscope`. For example:
```html
<html>
<body>
<div id="product" itemscope itemtype="http://schema.org/Product" itemref="other-product-properties">
    <span itemprop="name">Executive Anvil</span>
    <img itemprop="image" src="img1.jpg"/>
</div>
<div id="other-product-properties">
    <img itemprop="image" src="img2.jpg"/>
    <div itemscope itemtype="http://schema.org/Product" itemprop="related_products">
        <span itemprop="name">REL PROD 1</span>
        <img itemprop="image" src="rel-prod-1.jpg">
    </div>
</div>
</body>
</html>
```

Is extracted as (see the duplicate name and the extra image):

```py
[{'@context': 'http://schema.org',
  '@type': 'Product',
  'image': ['img2.jpg', 'rel-prod-1.jpg', 'img1.jpg'],
  'name': ['REL PROD 1', 'Executive Anvil'],
  'related_products': {'@type': 'Product',
                       'image': 'rel-prod-1.jpg',
                       'name': 'REL PROD 1'}}]
```
This PR patchs the code to solve this problem, so that the extracted result is:

```py
[{'@context': 'http://schema.org',
  '@type': 'Product',
  'image': ['img2.jpg', 'img1.jpg'],
  'name': 'Executive Anvil',
  'related_products': {'@type': 'Product',
                       'image': 'rel-prod-1.jpg',
                       'name': 'REL PROD 1'}}]
```
I patched function `_extract_property_refs()` so that properties in a different scope than the referenced element are skipped. I have passed the tests and they pass.

### DIFF
--- a/extruct/w3cmicrodata.py
+++ b/extruct/w3cmicrodata.py
@@ -130,14 +130,16 @@ class LxmlMicrodataExtractor(object):
         if 'itemprop' in ref_node.keys() and 'itemscope' in ref_node.keys():
             # An full item will be extracted from the node, no need to look
             # for individual properties in childs
-            yield from extract_fn(ref_node)
+            for p, v in extract_fn(ref_node):
+                yield p, v
         else:
             base_parent_scope = ref_node.xpath("ancestor-or-self::*[@itemscope][1]")
             for prop in ref_node.xpath("descendant-or-self::*[@itemprop]"):
                 parent_scope = prop.xpath("ancestor::*[@itemscope][1]")
                 # Skip properties defined in a different scope than the ref_node
                 if parent_scope == base_parent_scope:
-                    yield from extract_fn(prop)
+                    for p, v in extract_fn(prop):
+                        yield p, v
 
     def _extract_property(self, node, items_seen, base_url):
         props = node.get("itemprop").split()
--- a/extruct/w3cmicrodata.py
+++ b/extruct/w3cmicrodata.py
@@ -79,17 +79,17 @@ class LxmlMicrodataExtractor(object):
                 item["id"] = itemid.strip()
 
         properties = collections.defaultdict(list)
-        # start with item references
+        for name, value in self._extract_properties(
+                node, items_seen=items_seen, base_url=base_url):
+            properties[name].append(value)
+
+        # process item references
         refs = node.get('itemref', '').split()
         if refs:
             for refid in refs:
                 for name, value in self._extract_property_refs(
                         node, refid, items_seen=items_seen, base_url=base_url):
                     properties[name].append(value)
-
-        for name, value in self._extract_properties(
-                node, items_seen=items_seen, base_url=base_url):
-            properties[name].append(value)
 
         props = []
         for (name, values) in properties.items():
--- a/extruct/w3cmicrodata.py
+++ b/extruct/w3cmicrodata.py
@@ -119,10 +119,21 @@ class LxmlMicrodataExtractor(object):
                 yield p, v
 
     def _extract_property_refs(self, node, refid, items_seen, base_url):
-        for prop in node.xpath("id($refid)/descendant-or-self::*[@itemprop]", refid=refid):
+        ref_node = node.xpath("id($refid)[1]", refid=refid)
+        if not ref_node:
+            return
+        ref_node = ref_node[0]
+        base_parent_scope = ref_node.xpath("ancestor-or-self::*[@itemscope][1]")
+        if 'itemprop' in ref_node.keys():
             for p, v in self._extract_property(
-                    prop, items_seen=items_seen, base_url=base_url):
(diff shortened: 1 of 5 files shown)


ivanprado:
I detected some cases not working... Work in progress.

ivanprado:
@lopuhin I'm finished with that. Feel free to change the reviewer if you think other person should be the reviewer, I just added you as it was the github suggestion.

tests/samples/schema.org/product-ref.json
```diff
@@ -0,0 +1,69 @@
...
+                "type": "http://schema.org/Product",
+                "properties": {
+                    "name": "REFERENCED PRODUCT",
+                    "image": "img-ref.jpg"
+                }
+            },
+            "prop3": "REFERENCED TO INCLUDE PROPERTIES AND ALSO INDIVIDUAL PRODUCT",
+            "image": [
+                "img-2.jpg",
+                "anvil_executive.jpg"
```
lopuhin: The order of the images seems reversed to the depth-first traversal order, I wonder if we should care about this.

tests/samples/schema.org/product-ref.json
```diff
@@ -0,0 +1,69 @@
...
+                    "itemCondition": "http://schema.org/UsedCondition",
+                    "availability": "http://schema.org/InStock"
+                }
+            }
+        }
+    },
+    {
+        "type": "http://schema.org/Product",
+        "properties": {
+            "prop3": "REFERENCED TO INCLUDE PROPERTIES AND ALSO INDIVIDUAL PRODUCT",
```
lopuhin: Sorry maybe I'm reading the schema incorrectly, but why do we have a separate product with its properties duplicated also in the product above?

ivanprado: The product with id `more-properties` does not have attribute `itemprop`. Because of that, when referenced only inner attributes are included. As the product has not been extracted when referencing, it is finally extracted as an individual Product.

Regarding if that should be behaviour or not, I'm not completelly sure. This is the standard definition of `itemref`https://www.w3.org/TR/microdata/#names:-the-itemprop-attribute but is not very complete. 

There is an algorithm definition here: https://www.w3.org/TR/microdata/#associating-names-with-items but I didn't go so deep for this patch as it seems it is not the same approach followed by `extruct`.

I have just tested this html in Google Dev Tools and it also extracts the product isolated, but it is not including `prop3` into the first product.

tests/samples/schema.org/product-ref.json
ivanprado: I'm also not sure. Looking at that: https://www.w3.org/TR/microdata/#associating-names-with-items I would say that referenced items should be processed after the current item. Maybe is just as easy as changing the order on which 
`itemref` is processed. I did a quick test and only one test fails... I'll check deeper later.

ivanprado: It makes sense to me to change the order in which `itemref` is processed. I have committed a fix.

extruct/w3cmicrodata.py
```diff
@@ -119,10 +121,25 @@ def _extract_properties(self, node, items_seen, base_url):
                 yield p, v
 
     def _extract_property_refs(self, node, refid, items_seen, base_url):
-        for prop in node.xpath("id($refid)/descendant-or-self::*[@itemprop]", refid=refid):
-            for p, v in self._extract_property(
-                    prop, items_seen=items_seen, base_url=base_url):
+        ref_node = node.xpath("id($refid)[1]", refid=refid)
```
lopuhin: As far as I understand, if there were several elements with the same ``id``, they were all extracted, but now just the first one is extracted? I realise that there supposed to be only one element with each ``id`` so this seems fine.

lopuhin:
@Kebniss @kmike I think this is ready to merge, would you like to have a look?

tests/samples/schema.org/product-ref.json
lopuhin: I see, thanks for checking this @ivanprado . I think the current approach makes sense for ``extruct``.

extruct/w3cmicrodata.py
ivanprado: I see... I went to standard: https://www.w3.org/TR/html401/struct/global.html#h-7.5.2

> id = name [CS]
> This attribute assigns a name to an element. This name must **be unique in a document**.

So I think is safe to spect just one.

Kebniss:
Looks good to me! Thank you for fixing this @ivanprado :)

lopuhin:
Thanks @ivanprado and @Kebniss

PR #115, opened 2019-06-07
Reverse priorities for repeated properties in uniform format for opengraph

ivanprado:
Some pages contain a duplicated definition of some properties like "og:image". See the following pages:
https://nerdist.com/article/star-wars-cast-reylo-episode-ix/
https://cleantechnica.com/2019/04/16/fukushimas-final-costs-will-approach-one-trillion-dollars-just-for-nuclear-disaster/

Extruct default behaviour seems to be keep the last one, meanwhile Facebook default behaviour seems to be keep the first one according to results at the developer tool (see https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fnerdist.com%2Farticle%2Fstar-wars-cast-reylo-episode-ix%2F for an example).

Extruct should mimic Facebook behaviour so this PR is reverting the priorities when flattening OpenGraph properties.

### DIFF
--- a/extruct/uniform.py
+++ b/extruct/uniform.py
@@ -4,7 +4,7 @@ from six.moves.urllib.parse import urlparse, urljoin
 def _uopengraph(extracted):
     out = []
     for obj in extracted:
-        flattened = dict(obj['properties'])
+        flattened = dict(reversed(obj['properties']))
         t = flattened.pop('og:type', None)
         if t:
             flattened['@type'] = t
--- a/tests/samples/songkick/elysianfields.html
+++ b/tests/samples/songkick/elysianfields.html
@@ -30,6 +30,7 @@
     <meta property="og:description" content="Buy tickets for an upcoming Elysian Fields concert near you. List of all Elysian Fields tickets and tour dates for 2017.">
     <meta property="og:url" content="http://www.songkick.com/artists/236156-elysian-fields">
     <meta property="og:image" content="http://images.sk-static.com/images/media/img/col4/20100330-103600-169450.jpg">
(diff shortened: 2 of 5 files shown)


ivanprado:
@lopuhin I see two Travis runs, one is succeeding and one is failing. I don't understand, what is the difference between both? (save for codecov)

lopuhin:
@ivanprado this looks like a genuine failure on python 3.4 (likely due to dictionary iteration order or something similar). Not sure what does this "push" build represent, looking...

lopuhin:
Aha this is the answer: https://stackoverflow.com/questions/41524861/travis-failed-push-but-passed-pr/41534988

lopuhin:
So it could be that the failure is random, and one build was less lucky

ivanprado:
I see now the problem. It seems that PyRDFA is returning repeated properties in an inconsistent ordering.

ivanprado:
@lopuhin thank you for your insights, very useful. It seems that the failure was caused because RDFa returns inconsistent ordering with duplicated properties. I have a filled an issue, created an xfail test and patched test so that is is not failing.

lopuhin:
Great, thanks for taking care of this issue @ivanprado 👍

kmike:
I think this is not the only thing we need to do, as it breaks extraction on some websites. It seems we need to take first non-empty result. E.g. on https://www.triganostore.com/tente-de-camping-raclet-bora-4.html there are two `og:description` values, the first one is empty. https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fwww.triganostore.com%2Ftente-de-camping-raclet-bora-4.html shows that a non-empty one is extracted.

Issue #113, opened 2019-05-27
Extruct returns incorrectly formatted description property


jakubwasikowski:
Seems that `extruct` incorrectly interprets description with included HTML tags from microdata.

See the below description extracted from URL https://www.monsterpetsupplies.co.uk/cat/cat-flea-tick/johnsons-4-fleas-cats-kittens:

```python
>>> import extruct
>>> import requests
>>> from w3lib.html import get_base_url
>>> r = requests.get('https://www.monsterpetsupplies.co.uk/cat/cat-flea-tick/johnsons-4-fleas-cats-kittens')
>>> base_url = get_base_url(r.text, r.url)
>>> data = extruct.extract(r.text, base_url=base_url)
>>> data['microdata'][0]['properties']['description']
"Johnsons 4 Fleas Cats & Kittens - 3 Treatment Pack, 6 Treatment PackFor use with Cats and Kittens over 4 weeks of age between 1 and 11kg.Johnson's 4fleas tablets are an easy to use oral treatment to kill adult fleas found on your pet.Effects on the fleas may be seen as soon as 15 minutes after administration.Between 95 - 100% of fleas will be killed off in the first six hours, but ALL adult fleas will be gone after a day.These tablets can be given directly to the mouth or may be mixed in a small portion f our pet's favourite food and given immediately. Administer a single tablet on an day when fleas are seen on your pet. Repeat on any subsequent day as necessary. Do not give more than one treatment per day.You may notice your pet scratching more than usual for the first half hour after administration; this is completely normal and caused by the fleas reacting to Johnson's 4Fleas tablets.While highly effective by themselves, 4Fleas is great when used as part of a programme to eliminate fleas and their larvae from both pets and their surroundings."
```
As it can be seen, there is a problem with formatting, like lack of space between "Pack" and "For" or between "11kg." and "Johnson's".

It turns out that the problem is not because of description property content per-se, because it looks correctly on the page source: 

```html
[...]

PR #119, opened 2019-07-15
Fix incorrectly formatted description property

jakubwasikowski:
The following has been done in this PR:
- Fixed issue https://github.com/scrapinghub/extruct/issues/113 with incorrectly formatted description property.
- Added new test case with website for which the issue occured.
- Fixed old test cases (the usage of `html_text` gets rid of weird new lines).
- I had to select minimal version because six in version 1.10.0 causes errors for python3.4 when installing.

The fix was based on the code pushed by @kmike in this PR: https://github.com/scrapinghub/extruct/pull/114. Thanks @kmike!


requirements.txt
```diff
@@ -7,5 +7,6 @@ requests
 rdflib
 rdflib-jsonld
 mf2py>=1.1.0
-six
+six>=1.11
```
ivanprado: @jakubwasikowski why did you updated six version? Maybe this is useful for @croqaz in #120

ivanprado:
@jakubwasikowski is the code ready to be reviewed or still is in WIP state?

requirements.txt
croqaz: Yes, I did exactly the same thing, but updated to 1.12 :D

jakubwasikowski: Sorry, forgot to mention it in description of PR. I had to select minimal version because six in version 1.10.0 causes errors for python3.4 when installing.

jakubwasikowski:
Hey @ivanprado, the code is ready to be review! Updated title of the PR :+1:

requirements.txt
croqaz: So when I added six>=1.12 it still crashes for Python 3.4.
But six>=1.11 works 🤦‍♂

jakubwasikowski: For 1.11 it works, but for 1.12 it doesn't? Weird :man_facepalming: Especially taking into account that 1.12 is the latest version.

extruct/w3cmicrodata.py
```diff
@@ -18,11 +18,32 @@
...
+    scripts=True,
+    javascript=False,  # onclick attributes are fine
+    comments=True,
+    style=True,
+    links=True,
+    meta=True,
+    page_structure=False,  # <title> may be nice to have
+    processing_instructions=True,
+    embedded=False,  # keep embedded content
+    frames=False,  # keed frames
```
ivanprado:
```suggestion
    frames=False,  # keep frames
```

extruct/w3cmicrodata.py
```diff
@@ -18,11 +18,32 @@
...
 
 import lxml.etree
+from lxml.html.clean import Cleaner
 from w3lib.html import strip_html5_whitespace
+import html_text
 
 from extruct.utils import parse_html
 
 
+# Cleaner which is similar to html_text cleaner, but is less aggressive
```
ivanprado: As far as I see, the difference between this cleaner and the html_text one is `embedded=False` and `frames=False`. It would be nice to include this in the comment and the reasoning about why. I imagine the reason is that we want to include frames and iframes content as well, right?

ivanprado:
Thanks @jakubwasikowski for this fix. Everything looks great. I left some comments to try to understand better why we use a different `Cleaner` than default html-text.

jakubwasikowski:
Thanks for your review @ivanprado!

@croqaz, would you like to take a look as you were mentioned and you're in reviewers now :smile:?
I first added Konstantin, but forgot that he is on holidays.

croqaz:
> @croqaz, would you like to take a look as you were mentioned and you're in reviewers now 😄?

Sure, I'll take a look today. I also have a PR that might need a bit of attention, so it's a good trade 😆

jakubwasikowski:
So would you like me to review something @croqaz :smile:? Add me as reviewer if you'd like to :+1:

jakubwasikowski:
Thanks for your review @ivanprado and @croqaz! :beers:

extruct/w3cmicrodata.py
```diff
@@ -182,7 +203,8 @@ def _extract_property_value(self, node, items_seen, base_url, force=False):
             return self._extract_textContent(node)
 
     def _extract_textContent(self, node):
-        return u"".join(self._xp_clean_text(node)).strip()
+        clean_node = cleaner.clean_html(node)
```
kmike: hey! I'm concerned about performance implications of this; we're copying & cleaning a tree many times for each page. Could you please run it on a large sample of pages, to see how bad is an impact?

jakubwasikowski: Sure, will do that :+1:

extruct/w3cmicrodata.py
```diff
@@ -182,7 +203,8 @@ def _extract_property_value(self, node, items_seen, base_url, force=False):
             return self._extract_textContent(node)
 
     def _extract_textContent(self, node):
-        return u"".join(self._xp_clean_text(node)).strip()
```
lopuhin: Looks like ``_xp_clean_text`` is not used any more, can it be removed?

jakubwasikowski: Hey @lopuhin! Didn't notice that. Sure, will remove it :+1:

Issue #116, opened 2019-06-07
RDFa ordering not preserved on duplicated properties


ivanprado:
When a property is repeated (i.e. on a page with multiple images annotates as `og:image`) RDFa return it as a list but is not preserving order. Preserving order is important as usually the first image is the most important. An example of page where this would be happening:

https://cleantechnica.com/2019/04/16/fukushimas-final-costs-will-approach-one-trillion-dollars-just-for-nuclear-disaster/

It seems difficult to solve it in extruct as the problem seems to present in `PyRdfa` library, and it is even happening in the online service: https://www.w3.org/2012/pyRdfa/Overview.html#distill_by_uri+with_options

Related with https://github.com/scrapinghub/extruct/pull/115 (I created an xfail test for that in this PR)

kmike:
fwiw, extruct is using https://github.com/RDFLib/rdflib, not pyRdfa. This could be related (not 100% sure): https://github.com/RDFLib/rdflib/issues/538.

croqaz:
What would be the implications to use SET instead or LIST in the RDFa values?
Eg: The values would be converted to sets of frozen Dicts:

```
"http://ogp.me/ns#image": [
                {
                    "@value": "http://images.sk-static.com/images/media/img/col4/20100330-103600-169450.jpg"
                },
                {
                    "@value": "http://images.sk-static.com/SECONDARY_IMAGE.jpg"
                }
            ],
```

ivanprado:
@croqaz the order is important for values. Usually, some pages annotate several images but the most important one (the one with the biggest resolution or more relevant) is the first one. If a set is used the order is lost. Instead, it would better to get consistent ordering by having values in the order they appear in the page. That's unfortunately not the case for RDFa.

croqaz:
Thank you for the explanation @ivanprado 👍

adityas114:
Hi, I would like to work on this. It would be great if you can assign me the issue.

I am thinking when the parsed output is returned from rdflib in rdfa.py, we can use the document object to fix the order in extruct itself. I have started working on it and I can raise a pull request soon.

ivanprado:
Hi @adityas114 . Nice if you can work on it :+1: . In order to contribute, just create a PR with the fix. There is no need to assign the issue.

adityas114:
@ivanprado I have created a PR (#139) for this issue. Please have a look and let me know if it requires any changes.

PR #120, opened 2019-07-16
Ignore empty OpenGraph props

### LINKED ISSUE #117: First non-empty result should be extracted in case of OpenGraph
https://github.com/scrapinghub/extruct/pull/115 was a step in a right direction (prefer first results), but it seems it is not the whole solution, as empty results should not be prioritized. 

E.g. on https://www.triganostore.com/tente-de-camping-raclet-bora-4.html there are two og:description values, the first one is empty. https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fwww.triganostore.com%2Ftente-de-camping-raclet-bora-4.html shows that a non-empty one is extracted.

croqaz:
Ignore empty OpenGraph properties or content.

Fixes #117

Upgraded six>=1.11 to make Python 3.4 tests pass.

This was also done in https://github.com/scrapinghub/extruct/pull/119

### DIFF
--- a/extruct/opengraph.py
+++ b/extruct/opengraph.py
@@ -30,8 +30,8 @@ class OpenGraphExtractor(object):
             namespaces.update(self.get_namespaces(head))
             props = []
             for el in head.xpath('meta[@property and @content]'):
-                prop = el.attrib['property']
-                val = el.attrib['content']
+                prop = el.attrib['property'].strip()
+                val = el.attrib['content'].strip()
                 if prop == '' or val == '':
                     continue
                 ns = prop.partition(':')[0]
--- a/extruct/opengraph.py
+++ b/extruct/opengraph.py
@@ -32,6 +32,8 @@ class OpenGraphExtractor(object):
             for el in head.xpath('meta[@property and @content]'):
                 prop = el.attrib['property']
                 val = el.attrib['content']
+                if prop == '' or val == '':
+                    continue
                 ns = prop.partition(':')[0]
                 if ns in _OG_NAMESPACES:
                     namespaces[ns] = _OG_NAMESPACES[ns]
--- a/tests/samples/songkick/elysianfields.html
+++ b/tests/samples/songkick/elysianfields.html
@@ -28,6 +28,7 @@
     <meta property="og:type" content="songkick-concerts:artist">
     <meta property="og:title" content="Elysian Fields">
     <meta property="og:description" content="Buy tickets for an upcoming Elysian Fields concert near you. List of all Elysian Fields tickets and tour dates for 2017.">
+    <meta property="og:description" content="" />
     <meta property="og:url" content="http://www.songkick.com/artists/236156-elysian-fields">
     <meta property="og:image" content="http://images.sk-static.com/images/media/img/col4/20100330-103600-169450.jpg">
     <meta property="og:image" content="http://images.sk-static.com/SECONDARY_IMAGE.jpg">
--- a/tests/samples/songkick/elysianfields.html
+++ b/tests/samples/songkick/elysianfields.html
@@ -27,6 +27,7 @@
     <meta property="og:site_name" content="Songkick">
     <meta property="og:type" content="songkick-concerts:artist">
     <meta property="og:title" content="Elysian Fields">
+    <meta property="og:title" content="  ">
     <meta property="og:description" content="Buy tickets for an upcoming Elysian Fields concert near you. List of all Elysian Fields tickets and tour dates for 2017.">
     <meta property="og:description" content="" />
     <meta property="og:url" content="http://www.songkick.com/artists/236156-elysian-fields">
--- a/tests/samples/songkick/elysianfields.json
+++ b/tests/samples/songkick/elysianfields.json
@@ -232,6 +232,9 @@
             "http://ogp.me/ns#description": [
                 {
                     "@value": "Buy tickets for an upcoming Elysian Fields concert near you. List of all Elysian Fields tickets and tour dates for 2017."
+                },
+                {
+                    "@value": ""
                 }
             ],
             "http://ogp.me/ns#image": [
--- a/tests/samples/songkick/elysianfields.json
+++ b/tests/samples/songkick/elysianfields.json
@@ -253,6 +253,9 @@
             "http://ogp.me/ns#title": [
                 {
                     "@value": "Elysian Fields"
+                },
+                {
+                    "@value": "  "
                 }
             ],
             "http://ogp.me/ns#type": [
--- a/tests/test_extruct.py
+++ b/tests/test_extruct.py
@@ -5,7 +5,6 @@ import unittest
 import pytest
 
 import extruct
-from extruct import SYNTAXES
 from tests import get_testdata, jsonize_dict, replace_node_ref_with_node_id
 
 
--- a/tests/test_extruct.py
+++ b/tests/test_extruct.py
@@ -16,9 +16,13 @@ class TestGeneric(unittest.TestCase):
         body = get_testdata('songkick', 'elysianfields.html')
         expected = json.loads(get_testdata('songkick', 'elysianfields.json').decode('UTF-8'))
         data = extruct.extract(body, base_url='http://www.songkick.com/artists/236156-elysian-fields')
-        # See test_rdfa_not_preserving_order()
-        del data['rdfa'][0]['http://ogp.me/ns#image']
-        del expected['rdfa'][0]['http://ogp.me/ns#image']
+        # Sorting the values here because RDFa is not preserving ordering on duplicated properties.
+        # See https://github.com/scrapinghub/extruct/issues/116
+        # Also see test_rdfa_not_preserving_order()
+        for rdf in data['rdfa']:
+            for key, pairs in rdf.items():
+                if ':' in key and isinstance(pairs, list):
+                    rdf[key] = sorted(pairs, key=lambda e: e["@value"], reverse=True)
         self.assertEqual(jsonize_dict(data), expected)
 
     @pytest.mark.xfail
--- a/extruct/_extruct.py
+++ b/extruct/_extruct.py
@@ -54,6 +54,7 @@ def extract(htmlstring,
     if errors not in ['log', 'ignore', 'strict']:
         raise ValueError('Invalid error command, valid values are either "log"'
                          ', "ignore" or "strict"')
+
     try:
         tree = parse_xmldom_html(htmlstring, encoding=encoding)
     except Exception as e:
@@ -65,6 +66,7 @@ def extract(htmlstring,
             return {}
         if errors == 'strict':
             raise
+
     processors = []
     if 'microdata' in syntaxes:
         processors.append(
@@ -95,6 +97,7 @@ def extract(htmlstring,
             ('rdfa', RDFaExtractor().extract_items,
              tree,
              ))
+
     output = {}
     for syntax, extract, document in processors:
         try:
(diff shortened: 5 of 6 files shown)


extruct/opengraph.py
```diff
@@ -32,6 +32,8 @@ def extract_items(self, document, base_url=None):
             for el in head.xpath('meta[@property and @content]'):
                 prop = el.attrib['property']
                 val = el.attrib['content']
+                if prop == '' or val == '':
```
ivanprado: Should `prop` or `val` be stripped to also consider multiple spaces/non-visible-characters as empty?

croqaz: Yes, I actually did that initially and then I removed the .strip() because I thought maybe we don't want to change the actual values?
Writing the code is easy, not sure what are the implications to the other libraries using this..

croqaz: But again, HTML should ignore extra spaces in general... So...
I'll add the .strip()

ivanprado: Ok!

tests/test_extruct.py
```diff
@@ -17,9 +16,10 @@ def test_all(self):
         body = get_testdata('songkick', 'elysianfields.html')
         expected = json.loads(get_testdata('songkick', 'elysianfields.json').decode('UTF-8'))
         data = extruct.extract(body, base_url='http://www.songkick.com/artists/236156-elysian-fields')
-        # See test_rdfa_not_preserving_order()
-        del data['rdfa'][0]['http://ogp.me/ns#image']
-        del expected['rdfa'][0]['http://ogp.me/ns#image']
+        for rdf in data['rdfa']:
```
ivanprado: Add a comment with the explanation about why you are sorting elements for rdfa. You can maybe add the link to the bug: See https://github.com/scrapinghub/extruct/issues/116

tests/samples/songkick/elysianfields.json
```diff
@@ -230,16 +230,19 @@
...
+                    "@value": ""
+                },
                 {
                     "@value": "Buy tickets for an upcoming Elysian Fields concert near you. List of all Elysian Fields tickets and tour dates for 2017."
                 }
             ],
             "http://ogp.me/ns#image": [
                 {
-                    "@value": "http://images.sk-static.com/images/media/img/col4/20100330-103600-169450.jpg"
+                    "@value": "http://images.sk-static.com/SECONDARY_IMAGE.jpg"
```
ivanprado: Changing this order will create a bug in function `test_rdfa_not_preserving_order`. Why do you have to change the order here?

ivanprado: I guess the reason is that it is because you are comparing with the sorting values... could you sort the values in the test function instead? That way `test_rdfa_not_preserving_order` would keep working as xfail case.

tests/test_extruct.py
croqaz: Ok, it's a good idea 👍

tests/samples/songkick/elysianfields.json
croqaz: I'm trying to sort the list in this failing function to fix it. I'm almost there.

ivanprado: @croqaz the function `test_rdfa_not_preserving_order` should be preserved as it is. It is an `xfail` function, which means it is a test that should be passing, but it is not. So don't sort on this function, sort it in `test_all`. 

Probably you understood well, but just in case :-)

croqaz: I see. Well, to fix the test_rdfa_not_preserving_order() all we need to do is:

```py
for rdf in data['rdfa']:
            for key, pairs in rdf.items():
                if ':' in key and isinstance(pairs, list):
                    rdf[key] = sorted(pairs, key=lambda e: e["@value"])
        self.assertEqual(jsonize_dict(data)['rdfa'], expected['rdfa'])
```

croqaz: `test_rdfa_not_preserving_order()` fails randomly as it is exactly now, with my changes or not, because the order of the duplicate values is random.
So it doesn't matter if the duplicate values are sorted correctly or not in the JSON file.
The function is still a good way to replicate the bug.

ivanprado: I see. Anyway, still, the order of the values for this case in `elysianfields.json` should be reversed. The first image should be `http://images.sk-static.com/images/media/img/col4/20100330-103600-169450.jpg` and the second image should be `http://images.sk-static.com/SECONDARY_IMAGE.jpg` as it was before, because this is the appearance order in the page.

This change will require some more code in the function `test_all` so that test passes, but this is the correct way. Otherwise `test_rdfa_not_preserving_order` would be incorrect, so wouldn't represent properly a case that should pass but is failing. (with current PR code `test_rdfa_not_preserving_order` represents a case that is failing and would also fail even in case #116 is fixed)

tests/samples/songkick/elysianfields.html
```diff
@@ -28,6 +28,7 @@
     <meta property="og:type" content="songkick-concerts:artist">
     <meta property="og:title" content="Elysian Fields">
     <meta property="og:description" content="Buy tickets for an upcoming Elysian Fields concert near you. List of all Elysian Fields tickets and tour dates for 2017.">
+    <meta property="og:description" content="" />
```
Gallaecio: Maybe you could have the tests also check that whitespace characters, like spaces, are now ignored as well?

tests/samples/songkick/elysianfields.json
croqaz: I see your point. I'm working on it.

tests/samples/songkick/elysianfields.html
croqaz: Thank you! That's a good idea.

croqaz:
I believe I adressed all the comments.

The Travis build seems to constantly fail with Python 3.4 at `"pip install -U tox codecov"`.

ivanprado:
@croqaz I have noticed that the same is happening to master, see: https://travis-ci.org/scrapinghub/extruct/builds/558857852?utm_source=github_status&utm_medium=notification

Digging deeper, the failure in the logs is:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/six-1.10.0.dist-info/METADATA'
```

I googled and this could be relevant: https://github.com/tox-dev/tox-travis/issues/76. Could you have a look?

croqaz:
Squashed all commits to have a clean history on merge.

ivanprado (approved):
Thank @croqaz for implementing that! I see everything right after the changes.

tests/samples/songkick/elysianfields.html
croqaz: @Gallaecio Do you want to take a look or I can just go ahead and merge this?
This is my first contribution to Extruct ;)

kmike:
In https://github.com/scrapinghub/extruct/issues/117 idea was to only drop empty values when deciding which value to choose, to be on a safe side. But maybe we can drop them always, as in this PR. @croqaz could you plese check that properties with empty values are always ignored, e.g. using https://developers.facebook.com/tools/debug/sharing/?

croqaz:
@kmike Just to make sure I understand this correctly, because I think I might have misunderstood how to solve the problem.

This is the problem:
```html
<meta property="og:description" content="" />
<blah-blah>... </blah-blah> ...
<meta property="og:description" content="My cool website" />
```
Duplicate OpenGraph with the same name, but one of them has an empty value.
In many cases, the empty value is the first, but it might be second just as well.

Solution:
1. Do we want to completely drop OpenGraph meta tags with empty value?
2. Do we want to keep the tags with empty value, but move them at the end of the values list, so they have a lower priority?

In this PR I implemented the first option.

croqaz:
Oh, Facebook's Debug tool only shows *one* of the OpenGraph meta tags.
It seems to be the most relevant.
But I don't think they care much about the order - in case of images, I'm not 100% they always take the first one - but I might be wrong, because I don't know how to find a website that has multiple images in a "random" order and check what Facebook would extract...

kmike:
The question is: if there is only a single property with an empty value, is it returned or dropped?

croqaz:
Ok, I understand now. I wonder if we can quickly craft a page like that and feed that to Facebook and see how it works?
Because I don't know how to find a page like that "out there".

Actually I might craft a page like that relatively easy.
Let me try.

croqaz:
For:

```
<meta property="og:description" content="" />
<meta property="og:image" content="" />
<meta property="og:title" content="" />
```
Facebook Sharing Debugger shows:

* Provided og:image URL, was not a valid URL
* og:title - Taken from the <title> tag
* og:description - empty

croqaz:
I also tested with:
```html
<meta property="og:image" content="" />
<meta property="og:image" content="/images/logo.png" />
```

And Facebook Sharing Debugger shows: "Provided og:image URL, was not a valid URL"
But also the logo is corretly identified and displayed as a main image.

So if the image is first and the empty is second, OR if the image is second and the empty is first, the Sharing Debugger still finds the correct image.

jakubwasikowski:
Hey guys! Is there something which prevents us from merging that fix?
Would be great to have it merged because it blocks a bit a PR in another project.

ivanprado:
My cent: Looking at http://ogp.me/ and https://developers.facebook.com/docs/sharing/webmasters/#markup I don't see any definition of a property that could contain an empty value and still have some meaning. It seems that properties cannot be used to label a page somehow. So I would include the removal of every single property without value from the output. What do you think the rest of reviewers?

tests/samples/songkick/elysianfields.json
croqaz: Ok, this is done, I forgot to mention.
All the meta tags are converted to JSON objects and the sorting of JSON in the tests matches the actual HTML.

croqaz:
@ivanprado You mean in this case of: `property="og:title" content=" "` or `property="og:description" content=""` both properties should be dropped from the result?

ivanprado:
@croqaz well, I have some doubts regarding the fix. By one side, I would say that it should be the user using `extruct` the one with the responsibility to filter empty values **(option 1)**. `extruct` should only expose the data in the proper order (appearance order), empty or not. 

But by the other side, I have not seen anything in the protocol definition pointing that is possible to create properties with empty values, so removing all of them could be reasonable (**option 2**). But the question is, it is the same a document with `property="og:description" content=""` than a document without any annotation? I'm not sure 100%.

The solution of cleaning empty values but only if there is at least one non-empty value for the duplicated property seems like a strange mix (**option 3**). But maybe could do the job:
- User don't have to deal with empty values by himself, as they are removed 
- Meaning could be retained in case of empty content

But is weird as the user still will have to deal with empty value for the particular case. It is a little bit strange. 

So in conclusión, I have not a clear view about what we should do, the three options make some sense. The prefered one by my side would be the option 1, but I'm also ok with option 2 and 3. @kmike @Gallaecio, what do you think?

kmike:
I have the same concerns as Ivan. For OpenGraph we return a single value for an attribute, and empty values affect prioritization in case of multiple meta tags, so doing a correct prioritization (non-empty over empty) is fixing a bug. 

Removing all empty values is a different thing; it solves the prioritization problem, but it is a larger change, with unclear outcome - it can be a right thing to do (if having an empty value is the same as not having tag at all), or it can be a data loss (if a presence of a tag with empty value is meaningful information). To decide we need to answer this question, and I'm not sure how to do it :) My suggestion was a tool by Facebook (as standard says nothing), and it *seems* to confirm that empty value is meaningful - if I'm reading @croqaz's results properly, it does show an empty description if a description is present, but empty - and shows nothing if description is absent.

PR #121, opened 2019-07-26
OpenGraph skipping empty for properties with values on flattening

ivanprado:
Solves https://github.com/scrapinghub/extruct/issues/117

### DIFF
--- a/extruct/uniform.py
+++ b/extruct/uniform.py
@@ -9,8 +9,8 @@ def _uopengraph(extracted):
         # Ensuring that never empty value is returned if there is a duplicated
         # property with non empty value
         non_empty_props = {k for k, v in properties if v and v.strip()}
-        flattened = dict((k, v) for k, v in properties
-                          if k not in non_empty_props or (v and v.strip()))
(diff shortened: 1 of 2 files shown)


extruct/uniform.py
```diff
@@ -4,7 +4,13 @@
 def _uopengraph(extracted):
     out = []
     for obj in extracted:
-        flattened = dict(reversed(obj['properties']))
+        # In order of appearance in the page
+        properties = list(reversed(obj['properties']))
+        # Ensuring that never empty value is returned if there is a duplicated
+        # property with non empty value
+        non_empty_props = {k for k, v in properties if v and v.strip()}
+        flattened = dict((k, v) for k, v in properties
```
jakubwasikowski: Nitpick: you can use `{k: v ...}` instead of `dict((k, v) ...)` here.

ivanprado: You are right! Changed.

ivanprado:
Thanks @jakubwasikowski ! Merged.

Issue #111, opened 2019-04-23
RecursionError('maximum recursion depth exceeded while calling a Python object',)


ivanprado:
Rest API not working for extruct 0.7.2

Response for requests are:
```
{
url: "https://nerdist.com/article/star-wars-cast-reylo-episode-ix/",
status: "error",
message: "RecursionError('maximum recursion depth exceeded while calling a Python object',)"
}
```

A warning is shown at startup: 
```
python -m extruct.service
/home/ivan/Documentos/scrapinghub/dev/extruct/extruct/service.py:3: MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. It may also silently lead to incorrect behaviour on Python 3.7. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016. Modules that had direct imports (NOT patched): ['urllib3.util (/home/ivan/Documentos/scrapinghub/dev/extruct/venv/lib/python3.6/site-packages/urllib3/util/__init__.py)', 'urllib3.util.ssl_ (/home/ivan/Documentos/scrapinghub/dev/extruct/venv/lib/python3.6/site-packages/urllib3/util/ssl_.py)']. 
  monkey.patch_all()
Bottle v0.12.16 server starting up (using GeventServer())...
Listening on http://0.0.0.0:10005/
Hit Ctrl-C to quit.
```

A possible solution could be in this message: https://github.com/gevent/gevent/issues/1235#issuecomment-395423091

pip list:
```
Package        Version    Location                                     
-------------- ---------- ---------------------------------------------
atomicwrites   1.3.0      
attrs          18.2.0     
beautifulsoup4 4.7.1      
bottle         0.12.16    
bumpversion    0.5.3      
certifi        2018.11.29 
chardet        3.0.4      
entrypoints    0.3        
extruct        0.7.2      
filelock       3.0.10     
flake8         3.7.5      
gevent         1.4.0      
greenlet       0.4.15     
html5lib       1.0.1      
idna           2.8        
isodate        0.6.0      
lxml           4.3.0      
mccabe         0.6.1      
mf2py          1.1.2      
more-itertools 5.0.0      
pip            10.0.1     
pluggy         0.8.1      
py             1.7.0      
pycodestyle    2.5.0      
pyflakes       2.1.0      
pyparsing      2.3.1      
pytest         4.2.0      
rdflib         4.2.2      
rdflib-jsonld  0.4.0      
[...]

Issue #63, opened 2018-03-21
Deprecate Server and HTTP Client mode?


cathalgarvey:
At present, `extruct` supports a HTTP API for "testing", but that carries a maintenance burden, and it invites feature-requests that may nudge it more and more into becoming a monolithic proxy service. That's not really where we want Extruct to be, I think.

Similarly with the HTTP-Client mode and the CLI tool that offers it - it's a mode of operation which probably shouldn't be our priority with `extruct`. I feel that if we provide a CLI client for `extruct`, it should probably just accept HTML through a Unix pipe or from a file, and operate on that. That way, people can use `curl` or `wget` or whatever else they like, and they won't worry about `extruct`'s support for various HTTP client features.

Thoughts? :)

stummjr:
@cathalgarvey I'm probably too biased to provide valuable feedback on this, as I've implemented the CLI in the past, but I think that it is an useful feature (at least I find myself using it use it frequently 😄).

But I totally agree with you on that `extruct` should just read stuff from a file/pipe, as it makes it easier to customize the requests as much as we need using a proper tool such as `curl`/`wget`.

IMHO, we should keep the CLI, but maybe change it so that `extruct` doesn't make the HTTP requests.

PR #125, opened 2020-01-08
Remove the HTTP API

Gallaecio:
Removes the HTTP API, as it adds complexity to the code base without adding much value.

Part 1 of 2 of https://github.com/scrapinghub/extruct/issues/63

Closes https://github.com/scrapinghub/extruct/issues/111


kmike:
Let's remove it - thanks @Gallaecio! 
If there is an interest, code can be easily moved out of extruct, to a separate repo.

PR #138, opened 2020-05-26
Added support for Open Graph arrays

### LINKED ISSUE #92: Add support for Open Graph Arrays
It is possible in Open Graph Protocol to specify more than one value for a single property. It's called OG Array http://ogp.me/#array. 

It seems that currently `extruct` doesn't support arrays in case the `uniform` option is set to `True`, because `uniform._uopengraph` function doesn't handle duplicated keys from the list of properties.

It'd be cool to add that support and return list in case there is more than one value or to just add some separate property with `list` suffix to be backward compatible.

ragnerok:
Fixes #92 
Once merged this would return Open Graph array when uniform is set True


extruct/uniform.py
```diff
@@ -1,21 +1,31 @@
...
+      non_empty_props = {k for k, v in properties if v and v.strip()}
+      # Set of repeated properties with at least 2 non empty values
+      repeated_props = {}
+      if with_og_arr:
+        repeated_props = {k for k in non_empty_props if len([i for i,v in properties if i==k and (v and v.strip())]) > 1}
+      # Add properties that is either duplicated but has only 1 non empty value
+      # or has only empty values
+      flattened = {k: v for k, v in properties
+                       if k not in repeated_props and (k not in non_empty_props or (v and v.strip()))}
+      if with_og_arr:
```
lopuhin, line 18: I think there are two options on how to make this change backwards-compatible, either add a prefix, or have a flag - I think if we have a flag like now, there is no need to add a prefix, and it's better to output a list instead of the original key here, although I'm not 100% sure which is the best API.

ragnerok: Just to clarify, if we end up using the flag are you suggesting that we not use the list suffix ? Also what do you mean by "it's better to output a list instead of the original key here" ?

ragnerok:
@lopuhin I am sorry my text editor was set to 2 spaces, I'll fix it. I'll add a few test cases as well.

extruct/uniform.py
lopuhin: > if we end up using the flag are you suggesting that we not use the list suffix? 

Right, exactly - with the flag we can change behaviour, and having result like ``"og:image": [...]`` looks more convenient to me, what do you think?

> it's better to output a list instead of the original key here

Here I meant the same, that since we're using a flag, we can use original key, for example ``"og:image"`` instead of ``"og:image_list"``.

ragnerok: Ohh okay, makes sense thanks. Should I make the necessary changes and push them ?

extruct/uniform.py
```diff
@@ -1,16 +1,28 @@
 from six.moves.urllib.parse import urlparse, urljoin
 
 
-def _uopengraph(extracted):
+def _uopengraph(extracted, with_og_arr=False):
     out = []
     for obj in extracted:
-        # In order of appearance in the page
```
lopuhin, line 7: I think this comment still makes sense - could you please restore it, unless there is a reason to remove it?

extruct/uniform.py
```diff
@@ -1,16 +1,28 @@
...
         properties = list(reversed(obj['properties']))
-        # Ensuring that never empty value is returned if there is a duplicated
-        # property with non empty value
+        # Set of non empty properties
         non_empty_props = {k for k, v in properties if v and v.strip()}
+        # Set of repeated properties with at least 2 non empty values
+        repeated_props = {}
+        if with_og_arr:
+            repeated_props = {k for k in non_empty_props
+                              if len([i for i, v in properties if i == k and (v and v.strip())]) > 1}
```
lopuhin, line 14: I think it's possible to be more efficient here, right now this is O(n**2) because we loop over all properties for all non-empty properties, while it should be possible to build a dict with repeated properties and their values in one go I think.

extruct/uniform.py
```diff
@@ -1,16 +1,28 @@
...
+        # Add properties that either have only empty values or are duplicated and
+        # have only 1 non empty value
         flattened = {k: v for k, v in properties
-                     if k not in non_empty_props or (v and v.strip())}
+                     if k not in repeated_props and (k not in non_empty_props or (v and v.strip()))}
```
lopuhin, line 18: I think we should not exclude values if they are in repeated properties, at least when the ``with_og_arr`` key is not set. Would you also mind adding a test which checks this?

extruct/uniform.py
lopuhin: yes, that would be perfect, thank you!

extruct/_extruct.py
```diff
@@ -134,7 +135,7 @@ def extract(htmlstring,
         for syntax, uniform, raw, schema_context in uniform_processors:
             try:
                 if syntax == 'opengraph':
-                    output[syntax] = uniform(raw)
+                    output[syntax] = uniform(raw, with_og_array)
                 else:
                     output[syntax] = uniform(raw, schema_context)
```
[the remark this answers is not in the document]
ragnerok: Right makes sense, so I'll have to change both `_uopengraph` and `_umicrodata_microformat` right ?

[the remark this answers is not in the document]
lopuhin: > I'll have to change both _uopengraph and _umicrodata_microformat right ?

I think no changes there are required, because we pass ``schema_context`` only to ``_umicrodata_microformat``, and ``with_og_array`` only to ``_uopengraph`` - I just suggested to make it more clear that these are different arguments, as opposed to different values of the same arguments, sorry that I was not clear.

extruct/uniform.py
ragnerok: When `with_og_array` is not set, then for a duplicated key we get the first property. However if `with_og_array` is set then we get a reversed list (i.e order of appearance is opposite to that on the page).

extruct/uniform.py
ragnerok: Okay yeah you're right this is very inefficient will change

extruct/uniform.py
ragnerok: Actually on line 11 `repeated_props` is set to be empty so if `with_og_arr` is not set this will work fine. I could add a test case to check this but I think [one](https://github.com/scrapinghub/extruct/blob/master/tests/test_uniform.py#L12) exists already

extruct/_extruct.py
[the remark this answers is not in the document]
ragnerok: Oh my bad I got it confused with `**kwargs`

extruct/uniform.py
```diff
@@ -1,16 +1,27 @@
...
-def _uopengraph(extracted):
+def _uopengraph(extracted, with_og_array=False):
     out = []
     for obj in extracted:
         # In order of appearance in the page
         properties = list(reversed(obj['properties']))
-        # Ensuring that never empty value is returned if there is a duplicated
-        # property with non empty value
+        # Set of non empty properties
         non_empty_props = {k for k, v in properties if v and v.strip()}
```
lopuhin, line 10: Super minor, but it seems that the code could be simplified if ``non_empty_props`` is removed and instead this condition is checked directly below, because we need one check per element in any case.

tests/test_uniform.py
```diff
@@ -27,6 +27,26 @@ def test_uopengraph(self):
...
+                    "@type": "songkick-concerts:artist",
+                    "og:title": "Elysian Fields",
+                    "og:description": "Buy tickets for an upcoming Elysian Fields concert near you. List of all Elysian Fields tickets and tour dates for 2017.",
+                    "og:url": "http://www.songkick.com/artists/236156-elysian-fields",
+                    "og:image": [ "http://images.sk-static.com/3.jpg",
+                                  "http://images.sk-static.com/2.jpg",
+                                  "http://images.sk-static.com/1.jpg"],
+                }]
+        body = get_testdata('misc', 'ogarray_test.html')
+        data = extruct.extract(body, syntaxes=['opengraph'], uniform=True, with_og_array=True)
```
lopuhin, line 47: Thanks for adding a test! Could you please also add a check how we handle the same input data with ``with_og_array``?

ragnerok: > Could you please also add a check how we handle the same input data with with_og_array?

Do you mean a check to see how this data is handled when `with_og_array` is False ?

extruct/uniform.py
ragnerok: Okay will handle this

tests/test_uniform.py
lopuhin: > Do you mean a check to see how this data is handled when with_og_array is False ?

Yes, exactly - the reason why I think it's valuable is that we could check that we take the first out of multiple values for an attributes, and it would help in case of future development here. Thank you!

ragnerok: Actually there already exists a [test](https://github.com/scrapinghub/extruct/blob/master/tests/test_uniform.py#L12) for multiple values for an attribute. It's just that here I needed more than 2 values otherwise it was failing codecov. I can add this test if you want me to.

lopuhin: I see, thank - probably you mean this https://github.com/scrapinghub/extruct/blob/master/tests/test_uniform.py#L30 ? I think it's enough, thank you!

ragnerok: Actually I've added the test that you mentioned with setting`with_og_array=True` as well. This would make that additional file that I added useless so I've removed that.

extruct/uniform.py
```diff
@@ -1,16 +1,28 @@
...
+        properties = list(obj['properties'])
+        flattened = {}
+
+        for k, v in properties:
+            if k not in flattened.keys():
+                flattened[k] = v
+            elif v and v.strip():
+                # If og_array isn't required add first non empty value
+                if not with_og_array:
+                    flattened[k] = flattened[k] if flattened[k] and flattened[k].strip() else v
```
Gallaecio, line 17: :lipstick: 

```suggestion
                    if not flattened[k] or not flattened[k].strip():
                        flattened[k] = v
```

lopuhin:
Thanks @ragnerok and thanks for review @Gallaecio :+1:

Issue #45, opened 2017-06-11
Error parsing Json-Ld in  JsonLdExtractor


giordand:
When the JsonLdExtractor tries to parse json ld in some web page raise `ValueError; no json object could be decoded`. 
My solution was to catch the error in `JsonLdExtractor._extract_items(self, node) ` (because maybe the extractor detected some microdata or rdfa in the webpage but the error only occurs with json-ld, and if we catch the error in extruct.extract we'll lose that data) and by default return an empty list:

```python
def _extract_items(self, node):
        try:
            data = json.loads(node.xpath('string()'))
            if isinstance(data, list):
                return data
            elif isinstance(data, dict):
                return [data]
        except Exception as e:
            print e
        return []
```

redapple:
Hi @giordand , thanks for the report.
Would you happen to have an example URL where this happens?
I have a local stashed change that I had to apply for examples you get from schema.org,
the root cause being for me that there was some HTML comment at the start of the `script` text node
```
$ git diff HEAD 
diff --git a/extruct/jsonld.py b/extruct/jsonld.py
index fe555db..84dfe74 100644
--- a/extruct/jsonld.py
+++ b/extruct/jsonld.py
@@ -4,6 +4,7 @@ JSON-LD extractor
 """
 
 import json
+import re
 
 import lxml.etree
 import lxml.html
@@ -24,7 +25,12 @@ class JsonLdExtractor(object):
                          if item]
 
     def _extract_items(self, node):
-        data = json.loads(node.xpath('string()'))
+        script = node.xpath('string()')
+        try:
+            data = json.loads(script)
+        except ValueError:
+            # sometimes this is due to some JavaScript comment
+            data = json.loads(re.sub('^(\s*//.*)|(\s*<!--.*-->\s*)', '', script))
         if isinstance(data, list):
             return data
         elif isinstance(data, dict):
```

giordand:
Yes, i remember that in my case there was HTML comments too , so it should be fixed when you commit & push that changes. Let me ask you a question , when you commit that changes will it be available with a pip update command to the extruct library?

redapple:
I'll need to release a new version of extruct for the change to be available directly from PyPI via pip.
Note that pip also allows installing from git specific commits

redapple:
@giordand , it would be most helpful if you can provide a real example of a URL (or the HTML of it) where extruct failed, just to check if my patch really does solve your issue.

giordand:
[...]

PR #137, opened 2020-05-11
feat: add parser for JSON with JS comment

Kiollpt:
I encountered the issue of parsing `json-ld` with JS comments. Finally, I found the `jstyleson` package that could solve.

### Reproduce issue:
```
1. comment the key-value pair 
2. comment occurs after the key-value pair 

{
 '\t//"interactionCount": "", // 文章互動數 \r\n'
'\t\t"height": "1000", // 高度， 有預設或是的高度\r\n'
}
```

PR #141, opened 2020-05-29
Solves issue #45

bhavya17037:
Add jsonStringFixer.py, which has a function to add quotes around any required
text in a json string. Used this in jsonld.py to handle invalid jsonld string.


Gallaecio:
I wonder if https://github.com/scrapinghub/extruct/pull/137 already handles this.

PR #144, opened 2020-07-27
new fallback for extracting json

### LINKED ISSUE #143: https://lubelska.co.uk/JSONDecodeError: Extra data: line 21 column 1 (char 572) for URL
Seems that the issue is that the JSON-LD document is:
```
// <![CDATA[
{
  "@context": "http:\/\/schema.org\/",
  "name": "Lubelska",
  "@type": "Organization",
  "logo": "https://lubelska.co.uk/wp/wp-content/uploads/2019/05/Lubelska-1.jpg",
  "url": "https://lubelska.co.uk/",
  "sameAs": [
    "https://twitter.com/EdwardHowey",
    "https://www.facebook.com/Lubelska-309144763268698/",
    "https://www.pinterest.co.uk/lubelskaltd/",
    "https://www.instagram.com/lubelska1/"
  ],
  "contactPoint": [{
    "@type": "ContactPoint",
    "telephone": "+44 20 3911 5526",
    "email": "info@lubelska.co.uk",
    "contactType": "sales"
  }]
}
// ]]&gt;
```

and after the replacing in `jsonLd._extractItems()`:
```
            # sometimes JSON-decoding errors are due to leading HTML or JavaScript comments
            data = json.loads(
                HTML_OR_JS_COMMENTLINE.sub('', script), strict=False)
```

it becomes:
```
{
  "@context": "http:\/\/schema.org\/",
  "name": "Lubelska",
  "@type": "Organization",
  "logo": "https://lubelska.co.uk/wp/wp-content/uploads/2019/05/Lubelska-1.jpg",
  "url": "https://lubelska.co.uk/",
  "sameAs": [
    "https://twitter.com/EdwardHowey",
    "https://www.facebook.com/Lubelska-309144763268698/",
    "https://www.pinterest.co.uk/lubelskaltd/",
    "https://www.instagram.com/lubelska1/"
  ],
  "contactPoint": [{
    "@type": "ContactPoint",
    "telephone": "+44 20 3911 5526",
    "email": "info@lubelska.co.uk",
    "contactType": "sales"
  }]
}
// ]]&gt;
```

and naturally this part which was not replaced:
```
// ]]&gt;
```
causes the error.

Currently, having a badly formatted information tag causes the entire library to fail and return no data. Is this the intended behavior?

Vitiell0:
Fixes #143

### DIFF
--- a/extruct/jsonld.py
+++ b/extruct/jsonld.py
@@ -10,11 +10,10 @@ import lxml.etree
 
 from extruct.utils import parse_html
 
-HTML_OR_JS_COMMENTLINE = re.compile(r'^\s*(//.*|<!--.*-->)')
-
 
 class JsonLdExtractor(object):
-    _xp_jsonld = lxml.etree.XPath('descendant-or-self::script[@type="application/ld+json"]')
+    _xp_jsonld = lxml.etree.XPath(
+        'descendant-or-self::script[@type="application/ld+json"]')
 
     def extract(self, htmlstring, base_url=None, encoding="UTF-8"):
         tree = parse_html(htmlstring, encoding=encoding)
@@ -27,15 +26,42 @@ class JsonLdExtractor(object):
             if items for item in items if item
         ]
 
+    def _extract_json_objects(self, text, decoder=json.JSONDecoder()):
+        """Find JSON objects in text, and yield the decoded JSON data
+
+        Does not attempt to look for JSON arrays, text, or other JSON types outside
+        of a parent JSON object.
+
+        """
+        pos = 0
+        while True:
+            match = text.find('{', pos)
+            if match == -1:
+                break
+            try:
+                result, index = decoder.raw_decode(text[match:])
+                yield result
+                pos = match + index
+            except ValueError:
+                pos = match + 1
+
     def _extract_items(self, node):
         script = node.xpath('string()')
+
         try:
             # TODO: `strict=False` can be configurable if needed
             data = json.loads(script, strict=False)
         except ValueError:
             # sometimes JSON-decoding errors are due to leading HTML or JavaScript comments
-            data = json.loads(
-                HTML_OR_JS_COMMENTLINE.sub('', script), strict=False)
+            data = []
+            try:
+                text = str(script)
+                for result in self._extract_json_objects(text):
+                    if result:
+                        data.append(result)
+            except:
+                return None
+
         if isinstance(data, list):
             return data
         elif isinstance(data, dict):


Gallaecio:
Could you cover your changes with tests?

Vitiell0:
@Gallaecio I believe all the tests are still passing? The new helper func `_extract_json_objects` doesn't have a test but it's pretty straightforward.

Gallaecio:
The point of tests is to avoid regressions. So, ideally, the changes should include a test that “proves” that your changes work.

It’s quite important for long-term maintenance, and I think specifically in this case adding tests makes a lot of sense. For instance, it makes sure that in the future we do not break support for the scenario that you are fixing here while trying to support a different scenario. And we are bond to see such change attempts, people can be very creative in the way they break JSON :slightly_smiling_face:

advance512:
lgtm

kennylajara:
I don't know if you are willing to merge this pull request as it seem to be here for around 4 month now, but this solve the issue that the current HTML parser cause with `ld+json` scripts that uses non-english characters. Example:

`{
	"@context": "http://schema.org",
	"@type": "NewsArticle",
	"headline": "We found an acute just here: á",
	"author": {
		"@type": "Person",
		"name": "Lastname with tilde: ñ"
	}
}`

I am pretty new to Python, so, I am not sure on how to implement the test on this language yet. If you can create and merge this pull-request it would be awesome.

lopuhin:
Thanks for feedback @kennylajara , although the PR still needs more work IMO, it's great to know that it resolves problems for you so that we know it's important.

It looks like the problem you encountered is different from the original issue, could you please share the URL where you have seen such markup?

kennylajara:
Sure @lopuhin . You can see an example here: `https://www.diariolibre.com/actualidad/gobierno-comenzara-a-entregar-el-bono-navideno-de-rd-1500-desde-este-11-de-diciembre-NG23168270`

lopuhin:
> The alternative solution is to wrap Extruct inside of a Try/Except, but this will ignore other important markup (like opengraph) while the code here will extract the opengraph and valid the part of the Json+ld markup.

Currently, this can be partly achieved with ``errors="ignore"`` or ``errors="log"`` argument to ``extruct.extract`` - at least it would continue other formats even if one of them fails.

kennylajara:
Got it, you are right. `errors="ignore"` is working for me now.

Is there any other documentation? I don't see those options on the `.README` file.

I just wanna add one more comment:

> > The alternative solution is to wrap Extruct inside of a Try/Except, but this will ignore other important markup (like opengraph) while the code here will extract the opengraph and valid the part of the Json+ld markup.
> 
> Currently, this can be partly achieved with `errors="ignore"` or `errors="log"` argument to `extruct.extract` - at least it would continue other formats even if one of them fails.

Yes, almost... `errors=ignore` helps me to keep the scraper running but I still find this pull request usefull as it will not completly ingore the `ld-json` metadata. This code partially extract the metadata, just ignore the item containing the error.

lopuhin:
> Is there any other documentation? I don't see those options on the .README file.

Only here for now: https://github.com/scrapinghub/extruct/blob/c1e48037469d83579da0d20128b49e591423d716/extruct/_extruct.py#L17-L43

>  I still find this pull request usefull as it will not completly ingore the ld-json metadata. This code partially extract the metadata, just ignore the item containing the error.

Right 👍

Issue #147, opened 2020-08-24
Slow microdata extraction


lopuhin:
Microdata extraction for one page takes around 190 s, the page itself is 7 MB uncompressed with 5.5k comments, so looks quite slow. URL is `http://jeursen.design/2015/10/prowood-gent-belgie/?replytocom=1961`, gzipped html attached. Extraction time is dominated by ``w3cmicrodata.py:65(get_docid)`` which is expensive (profiling screenshot below).

Exact invocation used:

```
    result = extruct.extract(
        html,
        syntaxes=['opengraph', 'microdata', 'json-ld', 'microformat'],
        errors='ignore',
        uniform=True,
        encoding='utf8')
```

profiling:

<img width="1267" alt="Screenshot 2020-08-24 at 11 30 49" src="https://user-images.githubusercontent.com/424613/91022146-6ad43480-e5fd-11ea-94cb-787d08977f8e.png">

html:

[jeursen.html.gz](https://github.com/scrapinghub/extruct/files/5116664/jeursen.html.gz)

lopuhin:
(looking into fixing this)

PR #148, opened 2020-08-24
Fast get_docid for microdata parser (fixes GH-147)

lopuhin:
Previously, get_docid had quadratic complexity, as for each node, it
counted it's position in the tree iterating over all previous nodes.
Here we fix this by computing get_docid in the same way for the first
element, and then incrementing it by one, which gives the same result
but has linear complexity and is much faster (e.g. bringing time from
190+ s to 6 s on example from GH-147).
Other discarded implementation options:
- using some other identifier for the node - this would break backwards
  compatibility, as this property was exposed and documented at least in
  the code
- instead of calling get_docid for the first element, just iterate over
  the whole document (from root) to build ids - this is less
  efficient in case we call extraction on a part of the document.

TODO:

- [x] make sure tests pass with an extra assert from 306bfb5
- [x] make sure that assert from 306bfb5 (replaced with sys.exit to avoid error being caught with errors="ignore") does not trigger on 1k real-world product items
- [x] revert 306bfb5

### DIFF
--- a/extruct/w3cmicrodata.py
+++ b/extruct/w3cmicrodata.py
@@ -65,7 +65,7 @@ class LxmlMicrodataExtractor(object):
 
     def get_docid(self, node, itemids):
         try:
-            return itemids[node]  # same as self.get_docid(node, {})
+            itemid = itemids[node]  # same as self.get_docid(node, {})
         except KeyError:
             # Even after itemids are built,
             # this might fail if extract_items is called on a part of the document,
@@ -73,6 +73,10 @@ class LxmlMicrodataExtractor(object):
             # although this does not look likely in practice,
             # so not a performance concern.
             return int(self._xp_item_docid(node))
+        else:
+            # will be removed, this is to make sure optimization is correct
+            assert itemid == self.get_docid(node, {})
+            return itemid
 
     def extract(self, htmlstring, base_url=None, encoding="UTF-8"):
         tree = parse_html(htmlstring, encoding=encoding)
--- a/extruct/w3cmicrodata.py
+++ b/extruct/w3cmicrodata.py
@@ -45,34 +45,18 @@ cleaner = Cleaner(
 
 
 class LxmlMicrodataExtractor(object):
-    # iterate in document order (used below for get_docid optimization)
+    # iterate in document order (used below for fast get_docid)
     _xp_item = lxml.etree.XPath('descendant-or-self::*[@itemscope]')
     _xp_prop = lxml.etree.XPath("""set:difference(.//*[@itemprop],
                                                   .//*[@itemscope]//*[@itemprop])""",
                                 namespaces = {"set": "http://exslt.org/sets"})
     _xp_clean_text = lxml.etree.XPath('descendant-or-self::*[not(self::script or self::style)]/text()')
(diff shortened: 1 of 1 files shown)


lopuhin:
hi @ivanprado could you please review? I think solution works well, but implementation is tricky, maybe there is some simpler approach here.

extruct/w3cmicrodata.py
```diff
@@ -62,23 +63,47 @@ def __init__(self, nested=True, strict=False, add_text_content=False, add_html_n
         self.add_text_content = add_text_content
         self.add_html_node = add_html_node
 
-    def get_docid(self, node):
-        return int(self._xp_item_docid(node))
+    def get_docid(self, node, itemids):
+        try:
+            return itemids[node]  # same as self.get_docid(node, {})
+        except KeyError:
+            # Even after itemids are built,
```
ivanprado, line 70: Alternative: compute always all ids starting from the root in `_build_itemids`. According to your comment, this will be the most common case, and for the case in which only a section of the document is extracted, I have the feeling that the speed would be equivalent or faster.  But probably code will simplify a little bit as `_xp_item_docid` wouldn't be required anymore.

ivanprado (commented):
@lopuhin I had a quick first look, and it looks right. Nice catch, it was pretty inefficient. 

As a suggestion, given that we have two compatible implementations - `_xp_item_docid` and using the dict from `_build_itemids` - I think it would be nice to have a unit test that checks that both implementations return equivalent results using the sample pages in the test folder.

extruct/w3cmicrodata.py
lopuhin: Yes this makes sense @ivanprado , let me try this approach - I discarded this initially but now I see that performance-wise it should be very similar and looks simpler.

lopuhin: Implemented in 6f2e2d2, looks much simpler indeed, thanks for suggestion @ivanprado

lopuhin:
Thanks for review and suggestions @ivanprado , could you please have another look?

> I think it would be nice to have a unit test that checks that both implementations return equivalent results using the sample pages in the test folder.

Now we have only one implementation, and backwards compatibility is ensured by having ``iid`` attribute in tests, I get 2 failures with the following diff:

```
-        return {node: idx + 1 for idx, node in enumerate(self._xp_item(root))}
+        return {node: idx + 2 for idx, node in enumerate(self._xp_item(root))}
```

```
============================================================================== short test summary info ==============================================================================
FAILED tests/test_microdata.py::TestMicrodataFlat::test_w3c_5_2 - AssertionError: Lists differ: [{'iid': 2, 'type': ['http://md.example.com/loco',[520 chars]']}}] != [{'iid': 1, ...
FAILED tests/test_microdata.py::TestMicrodataFlat::test_w3c_7_1 - AssertionError: Lists differ: [{'iid': 2, 'type': ['http://schema.org/BlogPostin[737 chars]']}}] != [{'iid': 1, ...
=========================================================================== 2 failed, 62 passed in 2.29s ============================================================================
```

ivanprado (approved):
Thank you @lopuhin . It looks great!

lopuhin:
Thank you for review and suggestions @ivanprado !

PR #150, opened 2020-09-01
Document how to make releases

lopuhin:
TODO:

- [x] try to follow this for 0.10.0 release

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -660,4 +660,13 @@ Use bumpversion_ to conveniently change project version::
     bumpversion minor  # 0.0.1 -> 0.1.0
     bumpversion major  # 0.1.0 -> 1.0.0
 
+You can use ``--dry-run --verbose`` flags first to preview changes.
+
 .. _bumpversion: https://pypi.python.org/pypi/bumpversion
+
+Making releases
+---------------
+
+* Update ``HISTORY.rst`` in master (via PR or directly)
+* use bumpversion (see above) to update the version
+* push changes with tags: ``git push --follow-tags``
--- a/README.rst
+++ b/README.rst
@@ -656,9 +656,9 @@ Versioning
 
 Use bumpversion_ to conveniently change project version::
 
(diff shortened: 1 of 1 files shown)


lopuhin:
Sorry for the mess with multiple tried to make a release in master, original sequence didn't work and then I didn't notice that upload already succeeded on the next try. Documented the working sequence, hope it goes smoother next time.

kmike:
Thanks @lopuhin! 
The sequence makes sense to me, especially as you checked that it works :)

Do you think we can move it to wiki? 
This is information for maintainers, not for general extruct users.

lopuhin:
> Do you think we can move it to wiki? This is information for maintainers, not for general extruct users.

@kmike good point, I can move it along with bumpversion mention. Another option besides the wiki is to have it in git, e.g. in ``RELEASE.rst`` or ``MAINTENANCE.rst`` for easier discoverability, what do you think?

kmike:
I think both are fine, not strong preference. I've seen @Gallaecio updating release procedure in wiki almost on each release sometimes, when it's complex, but here it looks pretty straightforward.

lopuhin:
Aha I didn't know we already use wiki for that purpose, nice, make sense to follow the same approach here.

kmike:
We're not using it for this repo, but we have release procedure in wiki for Scrapy, Splash and some other Open Source packages we maintain - see e.g. https://github.com/scrapy/scrapy/wiki/Scrapy-release-procedure or https://github.com/scrapinghub/splash/wiki/How-to-make-a-Splash-release.

lopuhin:
Thanks for context @kmike , makes sense to use a similar approach across our repos. Created a wiki page https://github.com/scrapinghub/extruct/wiki and removed stuff from the README.

kmike:
Thanks @lopuhin for taking care of that!

lopuhin:
thanks @kmike !

PR #152, opened 2020-09-07
Add "product" namespace support for opengraph

lopuhin:
This is not a standard namespace (not mentioned in https://ogp.me/), but it's quite common in the wild, e.g. in a product page dataset where OG is present on ~35% of pages, this "product" namespace is present on 3.4% of pages, or ~10% of OG pages, and can often contain new information.

``tests/samples/misc/opengrapg_test.json`` is removed as this is a typo, correct ``tests/samples/misc/opengraph_test.json`` is also present.


jakubwasikowski (approved):
Nice catch @lopuhin, I wasn't aware there is a namespace like that.
I don't have any suggestion re the PR. Looks great.
Approved :+1:

lopuhin:
Thanks @jakubwasikowski 👍

Issue #135, opened 2020-04-24
ImportError: No module named 'rdflib.plugins.parsers.pyRdfa'


svnshikhil:
This issue occurs when we install warcio package

from rdflib.plugins.parsers.pyRdfa import pyRdfa as PyRdfa, Options, logger as pyrdfa_logger
ImportError: No module named 'rdflib.plugins.parsers.pyRdfa'

lopuhin:
@svnshikhil this should be fixed in the lateest release, 0.9.0, could you please check it?

saggu:
@lopuhin there seems to be a version conflict re:rdflib

`ERROR: extruct 0.9.0 has requirement rdflib<5.0.0, but you'll have rdflib 5.0.0 which is incompatible.`

Will this be fixed?

Gallaecio:
Do you need rdflib 5.0.0?

saggu:
I don't but we are using other libraries as `SPARQLWRAPPER` which is using the latest rdflib.
For future proofing we use the latest version of libraries anyway.

lopuhin:
@saggu I see, unfortunately ``extruct`` is not compatible with latest rdflib 5.0.0, we put a restriction on the version in 0.9.0 but you are right that this is not a viable solution long term, let's use this issue to track this.

I think the plan to improve this could be:
1. at least don't fail to import with latest rdflib, so that we loos only extractions provided by it
2. move off rdflib, either to another lib or adding our implementation, potentially derived from rdflib

svnshikhil:
> @svnshikhil this should be fixed in the lateest release, 0.9.0, could you please check it?

Yeah, it's working 

Thank you all

lopuhin:
Re-opening because the proper fix is still not in place, we want to allow users of extruct to use the latest rdflib, see https://github.com/scrapinghub/extruct/issues/135#issuecomment-619824716

lopuhin:
This is fixed by #161 and released in v0.12.0

PR #161, opened 2020-12-08
Add support for rdflib 5.0.0

### LINKED ISSUE #131: ModuleNotFoundError: No module named 'rdflib.plugins.parsers.pyRdfa' for RDFLib v5.0.0
```
>>> import extruct
RDFLib Version: 5.0.0
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/.pyenv/versions/3.6.10/envs/project/lib/python3.6/site-packages/extruct/__init__.py", line 1, in <module>
    from ._extruct import SYNTAXES, extract
  File "/home/user/.pyenv/versions/3.6.10/envs/project/lib/python3.6/site-packages/extruct/_extruct.py", line 5, in <module>
    from extruct.rdfa import RDFaExtractor
  File "/home/user/.pyenv/versions/3.6.10/envs/project/lib/python3.6/site-packages/extruct/rdfa.py", line 14, in <module>
    from rdflib.plugins.parsers.pyRdfa import pyRdfa as PyRdfa, Options, logger as pyrdfa_logger
ModuleNotFoundError: No module named 'rdflib.plugins.parsers.pyRdfa'
```

jayaddison:
This changeset adds support for `rdflib` 5.0.0 to the `extruct` library.

Users of `extruct` currently often encounter issue #131 (module not found) when they upgrade to `rdflib` 5.0.0; this is due to the removal of the `pyRdfa` parser plugin as [noted here](https://github.com/scrapinghub/extruct/issues/135#issuecomment-619824716).

Fortunately, the standalone [`pyRdfa3` library](https://github.com/RDFLib/pyrdfa3/) added support to register as an `rdflib` plugin in https://github.com/RDFLib/pyrdfa3/pull/26.

What this means is that when `pyRdfa3` is installed in the Python environment, it should restore RDFA parsing functionality for users of `rdflib` 5.0.0 or greater, and resolve the `ModuleNotFoundError`.

The currently-published PyPi package [`pyRdfa3`](https://pypi.org/project/pyRdfa3/), version 3.5.3, includes this functionality via `rdf.plugins.parser` in `entry_points.txt`

**NB**: This release isn't currently tagged in the source repo, but I think `pyRdfa3` v3.5.3 is from around commit https://github.com/RDFLib/pyrdfa3/commit/1562b3f6d6d6749e462ef6780ebf8125127c7b6a or https://github.com/RDFLib/pyrdfa3/commit/b623cdd3853f8c314c71186b7af2e12ab74f37b9 based on looking at some diffs.

**NB**: A note of caution for review/usage is that full Python3 support has been added to `pyRdfa3` from version 4.0.0 (in https://github.com/RDFLib/pyrdfa3/pull/34) -- but from inspection of the Python-code-related changes, `extruct` does not appear to depend at all on the affected Python methods (notably `copyErrors` and `rdf_from_sources` which are both provided for [CGI-related context](https://github.com/RDFLib/pyrdfa3/pull/34/files?file-filters%5B%5D=.py#diff-0ad2f02c2014adb081f36398c2e45c891bc2a07302e396909a6afee6bd1f16eaL30) in the top-level [`processURI`](https://github.com/RDFLib/pyrdfa3/pull/34/files?file-filters%5B%5D=.py#diff-0ad2f02c2014adb081f36398c2e45c891bc2a07302e396909a6afee6bd1f16eaL720) method).

May fix #131 (alternative approach).


jayaddison:
:pause_button: Sigh, after all that, this upgrade might not be so simple after all.

When using `tox` to run tests, module resolution/plugin registration for `pyRdfa3` seemed to work fine.  But when running `py.test tests` the `ModuleNotFoundError` re-occurs.

jayaddison:
:play_or_pause_button: Ok, this is in progress again; the module resolution error was due to the fact that I'd failed to update some module imports within the library, and there was also an upstream change regarding the `lang` XML attribute that had to be reflected in the test expectations.

Tests pass under Python2.7 but I'm still seeing a test failure under Python3:

```
FAILED tests/test_rdfa.py::TestRDFa::test_wikipedia_xhtml_rdfa_no_prefix - AssertionError: Lists differ: ...
```

jayaddison:
The remaining issue with the `test_wikipedia_xhtml_rdfa_no_prefix` test is that the default `xmlns` attribute is serialized in a different order between Python 2.7 and Python 3.x environments.

Python 3.x's `minidom` implementation retains the user-supplied ordering for element attributes (https://github.com/python/cpython/pull/10219).

Python 2.7 sorts the attributes and as a result they meet the current `extruct` library test expectations.

XMLNS attributes are added by `pyRdfa` in [this block of code](https://github.com/RDFLib/pyrdfa3/blob/1562b3f6d6d6749e462ef6780ebf8125127c7b6a/pyRdfa/utils.py#L229-L234) which assigns prefixed namespaces first, and then adds the default (unprefixed) namespace last.

When Python 2.7 sorts these attribute keys it brings `xmlns` (the default namespace) entry ahead of any prefixed (`xmlns:`, which sorts lexically after `xmlns`) namespace entries.

Python 3.7 retains the attribute ordering and leaves the default `xmlns` namespace attribute at the end of the list.

**Possible paths forward**
To to resolve the behaviour so that `extruct` tests pass; various options are available:

- Update the test expectations so that they do not care about attribute order
- Update the library so that the Python 3.7 implementation sorts the attributes, maintaining consistency with 2.7 and the existing tests
- Update `pyRdfa` to modify the order in which namespace attributes are assigned, and then update this library once that change is available

jayaddison:
Ok, the latest commit pushes a change for which I'm able to get tests to pass for both Python 2.7 and Python 3.9.  Unfortunately since Travis builds may not be running at the moment, this may not be apparent from the pull request as-is.

The approach taken was to add a workaround in the test suite, to ignore differences in attribute ordering.  This was achieved by using [`lxml`'s `canonicalize` method](https://lxml.de/api/index.html).

The rationale behind this is:

- It is not easily possible to make Python 2.7 retain attribute ordering without messy monkey-patching or other questionable approaches
- It does not seem clear that adding re-ordering of attributes in `pyRdfa` -- even if user-configurable -- would be a 'correct' fix
- Retaining attribute ordering in Python 3.7 seems desirable: having scrapers that return values in the format they were published can be considered a feature in many situations
- Adding a workaround in a test suite that can be removed later seems relatively innocuous

tests/samples/misc/Portfolio_Niels_Lubberman.json
```diff
@@ -27,15 +27,15 @@
     ],
     "http://purl.org/rss/1.0/modules/content/encoded": [
       {
-        "@value": "<p xml:lang=\"en\" xmlns=\"http://www.w3.org/1999/xhtml\" xmlns:content=\"http://purl.org/rss/1.0/modules/content/\" xmlns:dc=\"http://purl.org/dc/terms/\" xmlns:foaf=\"http://xmlns.com/foaf/0.1/\" xmlns:og=\"http://ogp.me/ns#\" xmlns:rdfs=\"http://www.w3.org/2000/01/rdf-schema#\" xmlns:sioc=\"http://rdfs.org/sioc/ns#\" xmlns:sioct=\"http://rdfs.org/sioc/types#\" xmlns:skos=\"http://www.w3.org/2004/02/skos/core#\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema#\">Op deze vernieuwde website kunt u enkele van mijn projecten vinden, tevens kunt u lessen downloaden die ik heb gemaakt.</p>\n\n",
```
lopuhin, line 30: This is the commit https://github.com/RDFLib/pyrdfa3/commit/9c893ecd92c5dfcf7ab8f441eecda787ba13bef6 referenced in 98f22052dc7e151daba9e16ff5051163c0b8d35a

lopuhin (commented):
Thanks for the PR @jayaddison , I agree with your call re python 2.7 tests, and the changes look great! Let me have another look at it, but so far I don't see any issues.

jayaddison:
Thanks @lopuhin!

lopuhin:
thanks for the PR @jayaddison 👍 

The changes look good to me.

There are two issues, both unrelated to this PR:

- There are some package version combinations allowed by our setup.py which lead to test failures (e.g. ``mf2py==1.0.5``, and also old ``rdflib-jsonld`` with new ``rdflib``), both in master and in this PR. It's probably not practical to enumerate all working combinations, but maybe we could at least have one combination which is known to work (in requirements.txt) and also check that latest versions allowed by setup.py work as well. This is obviously a topic for another PR, and not a blocker here.
- We don't have working builds at the moment, we'll probably need to migrate off Travis to fix this - again this is unrelated to this PR but I'd like to have working builds before we merge anything. Sorry for this issue.

So let us try to fix the builds first.

lopuhin (approved):
tests passed in #165 , merging, thanks again @jayaddison

lopuhin:
Btw this PR also fixes tests for python 3.8 and 3.9 (which were broken in master).

jayaddison:
Hi @lopuhin - not urgent, but I was wondering whether you know when there might be an updated release of `extruct`?  (I'd quite like to upgrade some libraries that have versioning pins in place as a result of the `rdflib` 5.0.0 compatibility issue)

lopuhin:
hi @jayaddison good point, let me work on it this week - the only blocker is moving the release to github actions from travis, should not take long.

jayaddison:
@lopuhin That'd be great, but take your time; I didn't realize that'd require setup in CI.  Thanks for the update.

lopuhin:
@jayaddison it's released now, thanks again for your contributions 👍

jayaddison:
Brilliant, thanks again @lopuhin!

PR #162, opened 2020-12-09
Bundle README.rst as long-form package description

### LINKED ISSUE #151: Description missing on PyPI
Here is how it looks currently at https://pypi.org/project/extruct/, probably would make sense to put README contents if it renders well enough:

Screenshot 2020-09-01 at 21 13 43

jayaddison:
Resolves #151

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
     version=get_version(),
     description='Extract embedded metadata from HTML markup',
     long_description=get_readme(),
-    long_description_content_type='text/markdown',
+    long_description_content_type='text/x-rst',
     author='Scrapinghub',
     author_email='info@scrapinghub.com',
     maintainer='Scrapinghub',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -16,6 +23,8 @@ def get_version():
     name='extruct',
     version=get_version(),
     description='Extract embedded metadata from HTML markup',
+    long_description=get_readme(),
+    long_description_content_type='text/markdown',
```
lopuhin, line 27: Since README is in RST, I think the type should be different, according to https://packaging.python.org/guides/making-a-pypi-friendly-readme/#including-your-readme-in-your-package-s-metadata
```suggestion
    long_description_content_type='text/x-rst',
```

lopuhin (commented):
Thanks @jayaddison , that's a great addition, left a comment above.

setup.py
jayaddison: Good catch, thanks!

lopuhin (approved):
Looks great 👍

PR #164, opened 2020-12-12
Enable github actions

lopuhin:
Only run tests for now

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -14,6 +14,7 @@ jobs:
 
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: ['2.7', '3.5', '3.6', '3.7', '3.8', '3.9']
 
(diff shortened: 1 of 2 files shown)


.github/workflows/python-package.yml
```diff
@@ -0,0 +1,33 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -r requirements-dev.txt
+    - name: tox
+      run: |
+        tox -e `python -c "import sys; print('py' + ''.join(sys.version.split('.')[:2]))"`
```
kmike, line 33: I think  `tox -e py` does the same thing:

> The environment py uses the version of Python used to invoke tox.

see https://tox.readthedocs.io/en/latest/example/basic.html#a-simple-tox-ini-default-environments

lopuhin: Aha that's much nicer, thanks @kmike 👍

PR #168, opened 2020-12-23
Make a release on PyPI via github actions

lopuhin:
TODO: set up secrets.

PR #170, opened 2020-12-28
Fix README, make sure it renders

lopuhin:
If it does not render, we won't be able to publish a release on PyPI, see https://github.com/scrapinghub/extruct/pull/168#issuecomment-751741909

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -680,7 +680,7 @@ extract the metadata from it directly from the command line.
 Dependencies
 ++++++++++++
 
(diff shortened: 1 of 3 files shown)


jayaddison:
D'oh, it'd have been useful for me to double-check this during #162.  I guess GitHub's markup rendering is more lenient than PyPi.  Thanks @lopuhin!

lopuhin:
No worries, it wasn't rendering correctly for quite a while, good that it's now fixed and looks much prettier on PyPI.

PR #173, opened 2021-03-15
add quotes around extruct[cli]

grafst:
It only worked after I added quotes around the extruct[cli] --> 'extruct[cli]'


Gallaecio:
What error did you get? What operating system and command-line interpreter did you use? When using Linux and Bash, this does not seem necessary.

Gallaecio (approved):
[Looks like an issue for zsh](https://github.com/iterative/dvc.org/pull/434).

lopuhin:
Thanks @grafst and @Gallaecio 👍

PR #180, opened 2021-08-10
Fix linguist stats

platelminto:
The current project stats show that it is mostly HTML due to an incorrectly set `.gitattributes`. This pull request fixes this.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-tests/samples/* linguist-vendored
+tests/samples/** linguist-vendored


Gallaecio (approved):

.gitattributes
```diff
@@ -1 +1 @@
-tests/samples/* linguist-vendored
+tests/samples/**/*.html linguist-vendored
```
Gallaecio, line 1: Shouldn’t it be just `tests/samples/**`?

platelminto: Ah, hadn't seen there were other files too (though they don't get included in linguist by default). I will change it, makes more sense.

Gallaecio (approved):

lopuhin (approved):
Looks good, thanks @platelminto

lopuhin:
New stats look good

Screenshot 2021-08-11 at 13 15 41

PR #188, opened 2021-12-15
Checking PR 182 build

lopuhin:
I'd like to double-check fresh status of #182 build and make sure I'm getting the same errors locally.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,9 @@
 # pip install -r requirements.txt
 lxml
 requests
-rdflib>=6.0.1
+rdflib>=6.0.1; python_version>="3.7"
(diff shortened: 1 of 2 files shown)


BryceStevenWilley (approved):
Looks good to me! (though I'm not an expert at tox). 

I'd be okay with merging this over #182 and just closing that PR.

lopuhin:
Great, thank you @BryceStevenWilley - I only made a small change on top of your PR, following the discussion in https://github.com/scrapinghub/extruct/pull/182#discussion_r718165091 - it seems we'd be also fine with 6.0.0 for newer python, so we no longer disallow it.

@Gallaecio would you mind having another look?

PR #189, opened 2022-01-09
Ignore invalid jsonld elements on the page source.

naveen17797:


extruct/jsonld.py
```diff
@@ -28,14 +28,23 @@ def extract_items(self, document, base_url=None):
             if items for item in items if item
         ]
 
+    def _may_be_get_json(self, script):
+        try:
+            return json.loads(script, strict=False)
+        except Exception:
+            return False
```
lopuhin, line 35: minor, but to me it would make more sense to return ``None`` here, what do you think? Both ``None`` and ``False`` can be valid results of JSON parsing, but ``None`` looks like a more "neutral" value to use for error case.

extruct/jsonld.py
```diff
@@ -28,14 +28,23 @@ def extract_items(self, document, base_url=None):
...
+
     def _extract_items(self, node):
         script = node.xpath('string()')
-        try:
-            # TODO: `strict=False` can be configurable if needed
-            data = json.loads(script, strict=False)
-        except ValueError:
-            # sometimes JSON-decoding errors are due to leading HTML or JavaScript comments
```
lopuhin, line 37: I'd rather not remove this comment, I think it's useful

extruct/jsonld.py
```diff
@@ -28,14 +28,23 @@ def extract_items(self, document, base_url=None):
...
-            data = jstyleson.loads(HTML_OR_JS_COMMENTLINE.sub('', script), strict=False)
+        data = self._may_be_get_json(script)
+        # check if valid json.
+        if not data:
+            script = jstyleson.dispose( HTML_OR_JS_COMMENTLINE.sub('', script))
```
lopuhin, line 42: Nitpick - let's remove a an extra space here
```suggestion
            script = jstyleson.dispose(HTML_OR_JS_COMMENTLINE.sub('', script))
```

extruct/jsonld.py
```diff
@@ -28,14 +28,23 @@ def extract_items(self, document, base_url=None):
...
+            data = self._may_be_get_json(script)
+        # After processing check if json is still valid.
+        if not data:
+            return False
```
lopuhin, line 46: Nitpick - we don't use return value, as this is a generator, so ``return`` would make more sense
```suggestion
            return
```

lopuhin (commented):
Thanks @naveen17797 , the feature makes sense, I left some minor comments regarding the code, however I also have a more general question - previously, if we had invalid JSON, we'd crash or log it (depending on the error setting). Now, it would be silently ignored. What do you think about still having some logging if our last attempt at parsing the JSON fails, similar to this? https://github.com/scrapinghub/extruct/blob/547c8a05e137ed19ac087e344667be3fbccfa335/extruct/_extruct.py#L111

extruct/jsonld.py
naveen17797: fixed.

extruct/jsonld.py
naveen17797: Added the comment back to the code.

tests/samples/custom.invalid/JSONLD_valid_and_invalid.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <script type="application/ld+json">
+        { foo: bar}
```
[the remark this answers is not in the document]
naveen17797: yes, it should ignore this jsonld element since it is not valid.

extruct/jsonld.py
naveen17797: fixed.

extruct/jsonld.py
naveen17797: fixed

naveen17797:
> previously, if we had invalid JSON, we'd crash or log it (depending on the error setting). Now, it would be silently ignored. What do you think about still having some logging if our last attempt at parsing the JSON fails, similar to this? 

i have added the log statement

`logging.exception('Invalid jsonld element detected %s', script)`

naveen17797:
i understand if we merge this PR this would remove the ability to stop parsing the page with invalid jsonld elements, previously extruct will raise an exception and fail for jsonld, i am not sure how could i retain this behaviour.

may be i can pass errors argument from extract() function to jsonld extractor @lopuhin  ?

argaurav:
Is there a timeline on this issue. It would be good to get a fix for this issue.

PR #190, opened 2022-01-14
Corrected typo in dublincore.py

susca:


lopuhin (approved):
Thanks for the catch @susca :+1: 

This is indeed a correct spelling: https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/publisher

lopuhin:
(tests are failing for unrelated reason, but merging #188 should fix that)

Issue #171, opened 2021-01-26
Crash on JSONDecodeError from body of YouTube page


wjdp:
I have some code to pull metadata from YouTube

```python
response = requests.get(video_url)
metadata = extruct.extract(response.text, base_url="https://youtube.com")
```

Have noticed some recent crashing, but only on some videos.

No crash: https://www.youtube.com/watch?v=ZY48KUAZKhM https://www.youtube.com/watch?v=ZlVI7YJGHq0
Crash: https://www.youtube.com/watch?v=987wzJ2NHBE https://www.youtube.com/watch?v=0-EF60neguk

Common factor among those that crash is apostrophes in the channel name!

```
Traceback (most recent call last):
  File "/home/will/local/breda/src/dredger/ingest/tests/test_youtube.py", line 72, in test_one
    youtube.get_video_data("https://www.youtube.com/watch?v=987wzJ2NHBE")
  File "/home/will/local/breda/src/dredger/ingest/youtube.py", line 46, in get_video_data
    metadata = extruct.extract(response.text, base_url="https://youtube.com")
  File "/home/will/.virtualenvs/breda/lib/python3.8/site-packages/extruct/_extruct.py", line 108, in extract
    output[syntax] = list(extract(document, base_url=base_url))
  File "/home/will/.virtualenvs/breda/lib/python3.8/site-packages/extruct/jsonld.py", line 25, in extract_items
    return [
  File "/home/will/.virtualenvs/breda/lib/python3.8/site-packages/extruct/jsonld.py", line 25, in <listcomp>
    return [
  File "/home/will/.virtualenvs/breda/lib/python3.8/site-packages/extruct/jsonld.py", line 38, in _extract_items
    data = jstyleson.loads(HTML_OR_JS_COMMENTLINE.sub('', script),strict=False)
  File "/home/will/.virtualenvs/breda/lib/python3.8/site-packages/jstyleson.py", line 123, in loads
    return json.loads(dispose(text), **kwargs)
  File "/usr/lib/python3.8/json/__init__.py", line 370, in loads
    return cls(**kw).decode(s)
  File "/usr/lib/python3.8/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.8/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Invalid \escape: line 1 column 211 (char 210)
```

[...]

PR #195, opened 2022-05-13
Solves issue #171

AbhinavSE:
As @wjdp suggested in the issue, an apostrophe in the channel's name causes the JSONdecode error. the json.loads() function fails when there are special characters like "\\x27" in the script. So I created a function in the jsonld extractor to replace the special characters before passing it into json.loads() function.

### DIFF
--- a/extruct/jsonld.py
+++ b/extruct/jsonld.py
@@ -30,6 +30,7 @@ class JsonLdExtractor(object):
 
     def _extract_items(self, node):
         script = node.xpath('string()')
+        script = self.clean_special_characters(script)
         try:
             # TODO: `strict=False` can be configurable if needed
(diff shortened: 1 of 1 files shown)


wjdp (approved):
Thanks @AbhinavSE, looks like this should work.

Unsure if there's a python lib that can handle the cleaning, do you expect characters outside the set provided? I'm thinking about special chars in other languages.

AbhinavSE:
> Thanks @AbhinavSE, looks like this should work.
> 
> Unsure if there's a python lib that can handle the cleaning, do you expect characters outside the set provided? I'm thinking about special chars in other languages.

There's less chances for there to be new characters, but these are YouTube channel names, so there's a chance. But for now, this should work without affecting the runtime by too much.

Issue #192, opened 2022-04-13
Some websites put meta tags outside the head.


paul-rchds:
On some pages meta tags are included outside of the head tag. For example on the YouTube channel page: https://www.youtube.com/c/Freecodecamp

As the opengraph extractor only looks in the head tag, all the og:* meta properties are missed.
In my fork, I changed the extractor to look in the body rather. 

If I get permission, I can do a PR?

Here is a link to where I made the change: https://github.com/scrapinghub/extruct/blob/c2cffbed26ae4ab8dd35d1860bfda00c3bac5783/extruct/opengraph.py#L28

lopuhin:
hi @paul-rchds yes, that would be great - I noticed the same issue myself but didn't get to implement everything required, here is a link to a PR https://github.com/scrapinghub/extruct/pull/129/files - feel free to start a new one.

frostrot:
I have changed the functionality of the extract_item function in OpengraphExtractor class, to incorporate the meta tags outside of the head. Have tested it on the link shared by @paul-rchds . Please review my PR for its workability. Thanks

PR #197, opened 2022-05-13
changed the opengraph meta data extraction to incorporate the html body.

frostrot:
#192 Added the feature to incorporate all the meta tags outside of the html head, by changing in the function extract_items() in class openClassExtractor. Furthermore, added a test case to named opengraph_test_2 which uses the html of [https://www.youtube.com/c/Freecodecamp](url) where the meta tags are also present in the body of the html, and the function is able to correctly identify all the tags and parse it.


lopuhin (commented):
Thanks for the PR @frostrot , that's a nice improvement, and thanks for adding tests. Still have a few things left to review. However, would you mind reverting changes unrelated to this PR, such as changing single quotes to double quotes, etc.? This would make it easier to track when and why certain parts of the code were changed.

lopuhin:
@frostrot also would you mind checking test failures? They look to be related to this PR.

Issue #199, opened 2022-10-01
[suggestion] adding type hints?


sbdchd:
I see this package supports python 2.7 so the type hints would have to be the comment variety:

https://web.archive.org/web/20220213164145/https://mypy.readthedocs.io/en/stable/cheat_sheet.html

Essentially:

```python
def add(a, b):
    # type: (int, int) -> int
    return a + b
```

instead of:

```python
def add(a: int, b: int) -> int:
    return a + b
```

lopuhin:
I think dropping python 2.7 support won't be a problem

sbdchd:
would you be open to a PR introducing type hints & dropping 2.7 support?

lopuhin:
Yes that would be helpful 👍 If you plan to make any code changes while dropping 2.7 support (e.g. clean-up six imports and other stuff required only for python 2.7), then I think it would be better to do so in a separate PR, and then add hints.

Regarding python 3 versions, I think we can definitely drop 3.5 support. In case 3.6 support would be a problem, we can discuss dropping that as well.

sbdchd:
Also, any thoughts on adding something like `black` for auto formatting? makes writing the types a little less cumbersome from the formatting changes

lopuhin:
Re black, we use it in a few projects, such as https://github.com/scrapinghub/web-poet/ and https://github.com/scrapinghub/scrapy-poet, and the folks running these projects are happy with the setup (cc @BurnzZ ). Do you have a preferred approach or experience migrating other projects @sbdchd ?

BurnzZ:
It'd be cool if we also have **pre-commit** hooks available with `isort` and `flake8` as well. :) Ref: https://github.com/scrapinghub/web-poet/blob/master/.pre-commit-config.yaml

sbdchd:
yeah basically in my experience adding type hints changes a ton of formatting, so having black helps a bit -- plus I like the look of the output :D

PR #200, opened 2022-10-06
cut old python versions from ci

sbdchd:
In preparation for adding type hints, this PR removes python 2.7 and 3.5 from the CI matrix.

Removing 2.7 means we can use the builtin types instead of type comments.

rel: https://github.com/scrapinghub/extruct/issues/199


lopuhin (commented):
hi @sbdchd thanks for the PR, could you please also update tox.ini and setup.py?

lopuhin (approved):
Looks good, thanks @sbdchd 👍

PR #201, opened 2022-10-10
add `pre-commit` setup with black and sort

sbdchd:
Some more setup for https://github.com/scrapinghub/extruct/issues/199

once merged we can add the commit sha to the ignored revs file

PR #202, opened 2022-11-21
create .git-blame-ignore-revs to contain commits due to black and isort

BurnzZ:
Follow-up on https://github.com/scrapinghub/extruct/pull/201


lopuhin (approved):
Looks good, thanks @BurnzZ 👍

Issue #215, opened 2024-03-31
Package breaking due to change in lxml


marcosfelt:
With new versions of lxml (I believe above 5.1.0), this package breaks with the following error:

```bash
ImportError: cannot import name '_ElementStringResult' from 'lxml.etree' 
```

I think the temporary solution is to pin the lxml version `>=5.0.0,<5.1.0`. I'm happy to submit a pull request for this.

PandaSanchez:
> With new versions of lxml (I believe above 5.1.0), this package breaks with the following error:
> 
> ```shell
> ImportError: cannot import name '_ElementStringResult' from 'lxml.etree' 
> ```
> 
> I think the temporary solution is to pin the lxml version `>=5.0.0,<5.1.0`. I'm happy to submit a pull request for this.

I also encountered this issue today and -- per @marcosfelt's recommendation -- pinning the lxml version fixed it. This saved the day for me, thanks!

jayaddison:
@marcosfelt @honzajavorek please double-check before making any adjustments based on this, but I believe that it was v5.1.1 (not v5.1.0) of `lxml` that introduced the relevant breaking change (https://github.com/lxml/lxml/commit/eba79343d0e7ad1ce40169f60460cdd4caa29eb3).

PR #217, opened 2024-04-01
fix: Add Support for lxml >= 5.2.0

michael-genson:
lxml 5.1.0 removed _ElementStringResult which we use during dom traversal. This PR attempts to import _ElementStringResult as usual, but falls back to defining it explicitly if the import fails. This fixes https://github.com/scrapinghub/extruct/issues/215

I chose to attempt to import first to preserve existing behavior if an older version of lxml is installed.

Additionally, we use the lxml HTML cleaner, which is now in its own separate package. Due to this I've added the optional html_clean package to the requirements.txt. As noted below, this raises a warning if using lxml < 5.2.0, but otherwise doesn't impact older versions.

### DIFF
--- a/extruct/xmldom.py
+++ b/extruct/xmldom.py
@@ -15,7 +15,7 @@ except ImportError:
 
     class _ElementStringResult(bytes):  # type: ignore[no-redef]
         """
-        _ElementStringResult is removed in lxml >= 5.1.0,
+        _ElementStringResult is removed in lxml >= 5.1.1,
         so we define it here for compatibility.
         """
 
--- a/extruct/xmldom.py
+++ b/extruct/xmldom.py
@@ -9,6 +9,19 @@ from xml.dom.minidom import Attr, NamedNodeMap
 from lxml.etree import ElementBase, XPath, _ElementUnicodeResult, tostring
 from lxml.html import HtmlElementClassLookup, HTMLParser
 
+try:
+    from lxml.etree import _ElementStringResult
+except ImportError:
+
+    class _ElementStringResult(bytes):  # type: ignore[no-redef]
+        """
+        _ElementStringResult is removed in lxml >= 5.1.0,
+        so we define it here for compatibility.
+        """
+
+        def getparent(self):
+            return self._parent  # type: ignore[attr-defined]
+
 
 class DomElementUnicodeResult:
     CDATA_SECTION_NODE = Node.CDATA_SECTION_NODE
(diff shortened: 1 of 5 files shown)


extruct/xmldom.py
```diff
@@ -47,7 +41,7 @@ def lxmlDomNodeType(node):
     if isinstance(node, ElementBase):
         return Node.ELEMENT_NODE
 
-    elif isinstance(node, (_ElementStringResult, _ElementUnicodeResult)):
+    elif isinstance(node, _ElementUnicodeResult):
```
jayaddison, line 44: This would require a baseline dependency of `lxml>=5.1.1`, but perhaps we could use `node.is_text` and `node.is_attribute` instead of the conditions within this code block?

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # project requirements, install them using following command:
 # pip install -r requirements.txt
-lxml
+lxml[html_clean]>=5.2.0
```
wRAR, line 3: `lxml[html_clean]` should be fine here, it will keep the older lxml version installed and just print a warning.

michael-genson: TIL, for some reason I thought this would raise an error

extruct/xmldom.py
michael-genson: I added a dummy class on import error since it doesn't exist in newer versions anyway, which should allow for full backwards compatibility

michael-genson:
mf2py [released v2 4 months ago](https://github.com/microformats/mf2py/releases/tag/v2.0.1) which has a single breaking change: it includes image alts in microdata by default (see changelog in release). The reasoning for this is in [this mf2py issue](https://github.com/microformats/mf2py/issues/182), but TL;DR the image alt has been part of the microformat spec for several years, and not including it was considered deprecated behavior (and now is removed entirely).

To accommodate this I've updated the tests, though it's not explicitly related to the lxml update.

wRAR (approved):

extruct/xmldom.py
```diff
@@ -6,15 +6,22 @@
...
+from lxml.etree import ElementBase, XPath, _ElementUnicodeResult, tostring
 from lxml.html import HtmlElementClassLookup, HTMLParser
 
+try:
+    from lxml.etree import _ElementStringResult
+except ImportError:
+
+    class _ElementStringResult(bytes):  # type: ignore[no-redef]
+        """
+        _ElementStringResult is removed in lxml >= 5.1.0,
```
jayaddison, line 18:
```suggestion
        _ElementStringResult is removed in lxml >= 5.1.1,
```

Ref: https://github.com/lxml/lxml/commit/eba79343d0e7ad1ce40169f60460cdd4caa29eb3

Gallaecio (approved):

PR #219, opened 2024-04-16
fix: remove python 3.7

Rotzbua:
## Changes

* Missing removals in #206
* add python 3.8 as requirement


Gallaecio (approved):

lopuhin (approved):
Thanks for all the cleanups @Rotzbua 👍

PR #226, opened 2024-04-17
feat: add `pyupgrade` to `pre-commit`

### LINKED ISSUE #221: chore: Remove Python 2 specific code
There are some outdated code from Python 2 and compatibility era. Should be removed.

- [x] utf8 encoding
- [x] compatibility imports
- [x] six dependency
- [x] `yield from` syntax
- [x] add `pyugprade` to pre-commit

Rotzbua:
## Changes

* update pre-commit dependencies
* add pyupgrade to avoid outdated python code / styles

## Reference

closes #221

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,16 +1,14 @@
 repos:
   - hooks:
       - id: black
-        language_version: python3
-    repo: https://github.com/ambv/black
-    rev: 22.10.0
+    repo: https://github.com/psf/black-pre-commit-mirror
+    rev: 24.4.0
   - hooks:
       - id: isort
-        language_version: python3
     repo: https://github.com/PyCQA/isort
-    rev: 5.12.0
+    rev: 5.13.2
   - hooks:
     - id: mypy
       additional_dependencies: [types-requests, types-mock, lxml-stubs]
     repo: https://github.com/pre-commit/mirrors-mypy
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -1,16 +1,18 @@
 repos:
   - hooks:
       - id: black
-        language_version: python3
-    repo: https://github.com/ambv/black
-    rev: 22.10.0
+    repo: https://github.com/psf/black-pre-commit-mirror
```
Gallaecio, line 4: TIL

Gallaecio (approved):

extruct/w3cmicrodata.py
```diff
@@ -74,7 +74,7 @@ def extract(self, htmlstring, base_url=None, encoding="UTF-8"):
 
     def extract_items(self, document, base_url):
         itemids = self._build_itemids(document)
-        items_seen: Set[Any] = set()
+        items_seen: set[Any] = set()
```
lopuhin, line 77: I'm curious how this works on Python 3.8 which we still support, I think this is a PEP 585 feature which was added in 3.9? https://peps.python.org/pep-0585/

lopuhin: Ah never mind I can see `from __future__ import annotations` above

kmike (approved):

PR #237, opened 2024-07-25
Add a minimal action I/O implementation to improve SearchAction support

Gallaecio:


extruct/w3cmicrodata.py
```diff
@@ -235,6 +235,17 @@ def _extract_property_value(self, node, items_seen, base_url, itemids, force=Fal
         elif node.get("content"):
             return node.get("content")
 
+        # https://schema.org/docs/actions.html#part-4
```
kmike, line 238: It seems most example in this link use an alternative representation, with a string (`"url-output": "required"`).

1. How is it handled in this PR, and how was it handled before? Could it be that we start to lose some data which was easily available before?
2. Do you know which approach is more common? Is it common on real websites to use the "Textual representations of Input and Output", or is it a rare edge case?

